PR #13, opened 2022-10-12
fix: build view-yaml bug

KristianKjerstad:
fix build error with view-yaml package
error with using the view-yaml when installing from npm/yarn - just get an error saying yaml plugin crashed. (found when trying to use the package in analysis portal)

### DIFF
--- a/packages/yaml-view/tsconfig.json
+++ b/packages/yaml-view/tsconfig.json
@@ -2,7 +2,7 @@
   "compilerOptions": {
     "outDir": "dist",
     "rootDir": "src",
-    "lib": ["dom", "dom.iterable", "esnext"],
+    "lib": ["dom", "dom.iterable", "commonjs"],
     "allowSyntheticDefaultImports": true,
     "forceConsistentCasingInFileNames": true,
     "resolveJsonModule": true,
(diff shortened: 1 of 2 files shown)


packages/yaml-view/tsconfig.json
```diff
@@ -1,6 +1,29 @@
...
+    "baseUrl": ".",
+    "declaration": true,
+    "declarationMap": true,
+    "sourceMap": true,
+    "allowJs": true,
+    "strict": true,
+    "skipLibCheck": true,
+    "jsx": "react",
+    "esModuleInterop": true,
+    "module": "esnext",
```
eoaksnes, line 21: @KristianKjerstad @soofstad  maybe this should be commonjs?

eoaksnes (changes requested):

netr0m (approved):
Approved, but see Eirik's remark

packages/yaml-view/tsconfig.json
KristianKjerstad: https://www.tsmean.com/articles/learn-typescript/typescript-module-compiler-option/

`"Using TypeScript, node.js and the TypeScript compiler option "module": "esnext" together is next to impossible. So if you're building code that should be ran with node.js, in 2020, you should still choose "module": "commonjs"."`

Seems like commonjs is recomended. Will update!

soofstad (approved):

PR #66, opened 2023-02-02
feature: header plugin

### LINKED ISSUE #65: dm-core-packages(dm-core): move header into separate plugin

soofstad:
- Adds the "header" plugin

- Provide an application header as a plugin

closes #65

NOTE: The attribute/grid selector is disabled in this first version of the plugin


packages/header/src/Header.tsx
```diff
@@ -0,0 +1,225 @@
...
+        {/*)}*/}
+      </TopBar.Header>
+      <TopBar.Actions>
+        <Icons>
+          <ClickableIcon onClick={() => setAboutOpen(true)}>
+            <Icon name="info_circle" size={24} title="About" />
+          </ClickableIcon>
+          <ClickableIcon onClick={() => setVisibleUserInfo(true)}>
+            <Icon name="account_circle" size={24} title="User" />
+          </ClickableIcon>
```
ingeridhellen, line 144: Shouldn't these only be shown if userInfo and about is set in THeaderPluginConfig?

packages/header/src/Header.tsx
```diff
@@ -0,0 +1,225 @@
...
+        </Icons>
+      </TopBar.Actions>
+      <Dialog
```
ingeridhellen, line 147: There is a TopBar.CustomContent that should be wrapped around this, I think

ingeridhellen, line 147: This syntax seems outdated. Perhaps update eds?

packages/header/src/Header.tsx
```diff
@@ -0,0 +1,225 @@
...
+        {/*    ))}*/}
+        {/*    <Link to={'/DMT/search'}>*/}
+        {/*      <AppBox>Search</AppBox>*/}
+        {/*    </Link>*/}
+        {/*  </AppSelectorWrapper>*/}
+        {/*)}*/}
+      </TopBar.Header>
+      <TopBar.Actions>
+        <Icons>
+          <ClickableIcon onClick={() => setAboutOpen(true)}>
```
ingeridhellen, line 139: I miss logic solving the issue of both dialogs being open at once.

packages/header/package.json
```diff
@@ -0,0 +1,32 @@
...
+    "dist",
+    "src",
+    "tsconfig.json"
+  ],
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "scripts": {
+    "prebuild": "rm -rf dist && yarn install",
+    "build": "tsc",
+    "postbuild": "cp -R blueprints package.json README.md ./dist/ && rm -rf node_modules"
```
ingeridhellen, line 30:
```suggestion
    "postbuild": "cp -R package.json README.md ./dist/ && rm -rf node_modules"
```

packages/header/src/useLocalStorage.tsx
```diff
@@ -0,0 +1,29 @@
+import { useState } from 'react'
+
+export function useLocalStorage<T>(
```
ingeridhellen, line 3: In mercury, this file has been adapted to include versioning of the local storage, making it much easier to give nice error messages if the data in someones local storage is no longer compatible with our code

ingeridhellen (changes requested):
I didn't succeed in testing this locally (running npm install && npm run build in all packages gave me several errors) so a bit hard to see how it will look

packages/header/src/Header.tsx
soofstad: Clicking outside the Dialog closes the first one automatically

packages/header/src/useLocalStorage.tsx
soofstad: IMO, this should be a generic localStorage hook. And any incompatible data stored there should be handled by which ever component is calling it.

packages/header/src/Header.tsx
soofstad: Updated EDS. Still seems correct. Also according to EDS docs.

soofstad: Wrapped with that. Don't think it changed anything :+1:

packages/header/src/Header.tsx
soofstad: fixed

packages/header/src/Header.tsx
```diff
@@ -0,0 +1,225 @@
...
+
+  if (isLoading || !entity) {
+    return <Loading />
+  }
+
+  return (
+    <TopBar>
+      <TopBar.Header>
+        {/*<ClickableIcon onClick={() => setAppSelectorOpen(!appSelectorOpen)}>*/}
+        <Icon name="grid_on" size={32} />
```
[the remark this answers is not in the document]
soofstad: Agreed. Ported as it was. But should use this instead.

packages/header/src/Header.tsx
```diff
@@ -0,0 +1,225 @@
...
+const UnstyledList = styled.ul`
+  margin: 0;
+  padding: 0;
+  list-style-type: none;
+  display: inline-flex;
+`
+
+type THeaderPluginConfig = {
+  userInfo: boolean
+  about: boolean
```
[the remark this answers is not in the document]
soofstad: fixed

packages/header/package.json
```diff
@@ -0,0 +1,32 @@
...
+  "dependencies": {
+    "@equinor/eds-core-react": "^0.20.0",
+    "@equinor/eds-icons": "^0.12.0",
+    "@equinor/eds-tokens": "^0.7.0",
+    "@development-framework/dm-core": ">=1.0.35"
+  },
+  "devDependencies": {
+    "@types/react": "^17.0.39",
+    "@types/styled-components": "^4.1.18",
+    "typescript": "^4.5.5"
```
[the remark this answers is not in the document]
soofstad: updated

packages/header/jest.config.js
```diff
@@ -0,0 +1,8 @@
+const base = require('../../../jest.config.base')
```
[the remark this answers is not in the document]
soofstad: fixed

eoaksnes (approved):

packages/header/src/useLocalStorage.tsx
ingeridhellen: It would still be a generic hook. https://github.com/equinor/mercury/pull/289/files

ingeridhellen: The problem is that js doesn't raise an error if the value returned by useLocalStorage does not fit with the type expected by the caller. Without versioning, the caller would have to check that the returned value, which could be quite complex, in fact has all the parameters expected by the type. Failure to add such a test when updating the type could cause failures anywhere in the code

PR #80, opened 2023-02-06
build: enable hot reloading of packages

eoaksnes:

* Enable hot reloading of packages 

* Makes it faster to develop packages and test them using the `examples` app

>***Note***
> To be able to get this to work a symlink from `dist/index.js` to `src/index.tsx` must exist. 
> I added a script to the examples/packages.json that will create and remove such symlinks.
> I might be some other way of solving this, but I did not find any better solution than symlinks but open to suggestions.


ingeridhellen (approved):
Can't say I understand all this, but it tested okay. Great work, this will help a lot!

PR #81, opened 2023-02-07
refactor: Standardize build scripts

ingeridhellen:

Standardizes the build scripts between packages
Ensures blueprints are uploaded to npm

### DIFF
--- a/packages/tabs/package.json
+++ b/packages/tabs/package.json
@@ -18,6 +18,7 @@
     "@development-framework/dm-core": ">=1.0.35"
   },
   "files": [
+    "blueprints",
     "dist",
     "src",
     "tsconfig.json"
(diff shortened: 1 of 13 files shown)


packages/tabs/package.json
```diff
@@ -18,15 +18,16 @@
     "@development-framework/dm-core": ">=1.0.35"
   },
   "files": [
+    "blueprints",
     "dist",
     "src",
```
soofstad, line 23:
```suggestion
```

soofstad (approved):

packages/tabs/package.json
ingeridhellen: src is needed to generate documentation: https://equinor.github.io/dm-docs/docs/libraries/@development-framework/dm-core/Components/Dialog. We can't get this from dist alone because comments are removed in the build step (I think, ask Morten for details)

PR #82, opened 2023-02-07
fix(core): create blueprint with name

### LINKED ISSUE #74: fix: "new blueprint" in explorer plugin creates a blueprint where name is missing
also, the created blueprint should not extend DefaultUiRecipes

soofstad:
- Add the user provided name of the blueprint before saving it

The "Create blueprint" dialog asks to provide a name. So it should be put in the new blueprint entity.

closes #74

### DIFF
--- a/packages/dm-core/package.json
+++ b/packages/dm-core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@development-framework/dm-core",
-  "version": "1.0.37",
+  "version": "1.0.38",
   "license": "MIT",
   "peerDependencies": {
     "react-router-dom": ">=5.1.2",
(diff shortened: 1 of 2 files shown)


eoaksnes (approved):
Nice work

KristianKjerstad (approved):
I think the explorer plugin also needs to be updated before https://github.com/equinor/dm-core-packages/issues/74 is closed (?)

also, the created blueprint should not extend DefaultUIRecipes. But this probably needs to be fixed in DMSS.

soofstad:
> I think the explorer plugin also needs to be updated before #74 is closed (?)
> 
> also, the created blueprint should not extend DefaultUIRecipes. But this probably needs to be fixed in DMSS.

Right, "explorer" need to use the latest "dm-core" version. Will fix.

Not sure what you mean by that last part...

PR #95, opened 2023-02-22
Feat/combine plugins into one

### LINKED ISSUE #91: refactor: group all plugins into dm-plugins

KristianKjerstad:
- Group all plugins into "dm-core-plugins"

- increase maintainability and "testability"

closes #91


KristianKjerstad:
NB!

I could not figure out why tests on the form package suddenly fails.
I had to do this change a lot of places:


```
expect(container.querySelector(`input[name="foo"]`))
```

->
```
expect(container.querySelector(`input[id="foo"]`))
```

Also, it seems like form package does not work with the TextareaWidget - see TextareaWidget.tsx
ALso, it seems like we are missing widget for number input (see changes in packages/dm-core-plugins/src/form/fields/NumberField.test.tsx)

KristianKjerstad:
Also, I could not make the mermaid plugin work. I think there is an issue with name conflicts with the "real" mermaid package https://www.npmjs.com/package/mermaid

packages/dm-core-plugins/package.json
```diff
@@ -0,0 +1,65 @@
...
+    "axios": "^0.27.2",
+    "golden-layout": "1.5.9",
+    "highlight.js": "^10.7.2",
+    "js-yaml": "^4.1.0",
+    "mermaid": "^10.0.0",
+    "react-contextmenu": "^2.11.0",
+    "react-copy-to-clipboard": "^5.1.0",
+    "react-notifications": "^1.4.3",
+    "styled-components": "^5.1.0",
+    "typescript": "^4.5.5",
```
ingeridhellen, line 19: this should be removed

packages/dm-core-plugins/package.json
```diff
@@ -0,0 +1,65 @@
...
+    "object-path": "^0.11.8",
+    "react-hook-form": "^7.31.2"
+  },
+  "devDependencies": {
+    "@development-framework/dm-core": ">=1.0.37",
```
ingeridhellen, line 24: shouldn't this be a normal dependency?

packages/dm-core-plugins/package.json
```diff
@@ -0,0 +1,65 @@
...
+    "jest": "^27.1.3",
+    "jest-environment-jsdom": "^27.1.1",
+    "react-router-dom": ">=5.1.2",
+    "react-test-renderer": "17.0.2",
+    "ts-jest": "^27.1.3",
+    "ts-loader": "^9.3.1",
+    "typescript": "^4.5.5",
+    "react": "17.0.2",
+    "react-dom": "17.0.2",
+    "styled-components": ">=5.1.0"
```
ingeridhellen, line 44: this is added to all three deps list. I suggest having it in the normal deps list only

packages/dm-core-plugins/package.json
```diff
@@ -0,0 +1,65 @@
...
+    "@types/styled-components": "^4.1.18",
+    "jest": "^27.1.3",
+    "jest-environment-jsdom": "^27.1.1",
+    "react-router-dom": ">=5.1.2",
+    "react-test-renderer": "17.0.2",
+    "ts-jest": "^27.1.3",
+    "ts-loader": "^9.3.1",
+    "typescript": "^4.5.5",
+    "react": "17.0.2",
+    "react-dom": "17.0.2",
```
ingeridhellen, line 43: Why lock the react versions?

packages/dm-core-plugins/src/explorer/components/context-menu/ContextMenu.tsx
```diff
@@ -245,7 +245,7 @@ export const NodeRightClickMenu = (props: TNodeWrapperProps) => {
               style={{ width: INPUT_FIELD_WIDTH }}
               type="string"
               value={formData?.name || ''}
-              onChange={(event) =>
+              onChange={(event: any) =>
```
ingeridhellen, line 248:
```suggestion
              onChange={(event: React.ChangeEvent<HTMLInputElement>) =>
```
perhaps?

packages/dm-core-plugins/src/explorer/components/context-menu/utils/SingleTextInput.tsx
```diff
@@ -17,7 +17,7 @@ export const SingleTextInput = (props: {
         <Input
           type={'string'}
           style={{ width: INPUT_FIELD_WIDTH }}
-          onChange={(event) => setFormData(event.target.value)}
+          onChange={(event: any) => setFormData(event.target.value)}
```
ingeridhellen, line 20:
```suggestion
          onChange={(event: React.ChangeEvent<HTMLInputElement>) => setFormData(event.target.value)}
```
perhaps?

packages/dm-core-plugins/src/index.tsx
```diff
@@ -0,0 +1,59 @@
+import { TPlugin } from '@development-framework/dm-core'
+import Editor from './explorer/Editor'
+import { PluginComponent } from './yaml-view/YamlPlugin'
+import { TabsContainer } from './tabs/TabsContainer'
+import { MermaidComponent } from './mermaid-plugin/MermaidComponent'
+import { JobInputEdit } from './job/JobInputEdit'
+import { JobControlWrapper } from './job/JobControlWrapper'
+import { FormComponent } from './form/FormComonent'
```
ingeridhellen, line 8:
```suggestion
import { FormComponent } from './form/FormComponent'
```
This is wrong in the filename too

packages/dm-core-plugins/src/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+import Header from './header/Header'
+import { DefaultPreviewComponent } from './default-preview/DefaultPreviewComponent'
+import { DefaultPdfComponent } from './default-pdf/DefaultPdfComponent'
+import { EditBlueprint } from './blueprint/EditBlueprint'
```
ingeridhellen, line 12: These names should perhaps be made more equal. Make a task for it?

packages/dm-core-plugins/src/mermaid-plugin/MermaidWrapper.js
```diff
@@ -0,0 +1,66 @@
...
+
+export default class MermaidWrapper extends React.Component {
+  // componentDidMount() {
+  //     mermaid.contentLoaded()
+  // }
+
+  render() {
+    // eslint-disable-next-line react/prop-types
+    // return <div className="mermaid">{this.props.chart}</div>
+    return <div>Mermaid is not working</div>
```
ingeridhellen, line 64: Has it never worked, or did this PR break it?

packages/dm-core-plugins/src/form/widgets/TextWidget/TextWidget.tsx
```diff
@@ -13,7 +13,16 @@ const TextWidget = (props: TWidget) => {
   const _onChange = ({
     target: { value },
   }: React.ChangeEvent<HTMLInputElement>) => onChange(value === '' ? '' : value)
-  return <TextField {...props} onChange={_onChange} label={label} />
+  return (
```
ingeridhellen, line 16: You'r sure we don't need the rest of the props?

ingeridhellen (changes requested):
Good work. Some small issues though

packages/dm-core-plugins/src/form/widgets/TextWidget/TextWidget.tsx
KristianKjerstad: ups, seems like I forgot to add inputRef, variant and helperText. Wil fix!

I could not use {...props} since the required namePath attribute on the props type TWidget is not a valid prop to TextField from EDS.

packages/dm-core-plugins/src/index.tsx
KristianKjerstad: good catch! Will fix

packages/dm-core-plugins/src/index.tsx
KristianKjerstad: yes, there is some need for more consistent naming and structure for the plugins.
Creating new issue for this

example/package.json
```diff
@@ -4,16 +4,10 @@
...
-    "@development-framework/form": "link:../packages/form",
-    "@development-framework/tabs": "link:../packages/tabs",
-    "@development-framework/header": "link:../packages/header",
-    "@development-framework/explorer": "link:../packages/explorer",
+    "@development-framework/dm-core-plugins": "link:../packages/dm-core-plugins",
     "@craco/craco": "^7.0.0",
-    "react": "^17.0.2",
-    "react-dom": "^17.0.2",
+    "react": "17.0.2",
+    "react-dom": "17.0.2",
```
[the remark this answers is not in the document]
KristianKjerstad: I experienced some strange error messages when running tests, and by googling it seemed like those errors could occur if using conflicting react versions. 

Double checked now, and there is no need to lock the react version. The errors when testing was caused by something else. Will fix!

packages/dm-core-plugins/package.json
KristianKjerstad: update: added styled-components to peer dependency, as recommended by Stig, but I had to add styled-components to dev dependencies also to make tests run.
Removed from normal dependencies list.

packages/dm-core-plugins/src/mermaid-plugin/MermaidWrapper.js
KristianKjerstad: Right now, mermaid plugin is not used in main (not imported in the example app):
image


Tried to add it back in main, and it also failed there.

 
 the mermaid plugin did work at some point (a few months ago). I experienced some strange dependency issues that I could not figure our

packages/dm-core-plugins/src/form/types.tsx
```diff
@@ -48,7 +48,7 @@ export type TBooleanFieldProps = {
   uiAttribute?: any
 }
 
-export declare type Variants = 'error' | 'default'
+export declare type Variants = 'error' | 'success' | 'error'
```
ingeridhellen, line 51:
```suggestion
export declare type Variants = 'error' | 'success' | 'warning'
```
as according to the eds docs https://eds-storybook-react.azurewebsites.net/?path=/docs/inputs-textfield--introduction

ingeridhellen (approved):

eoaksnes (approved):

PR #102, opened 2023-03-01
fix: tab label priority. label -> name -> attribute

### LINKED ISSUE #101: fix: tabs (sidebar) only uses "entity.name" for tabs text
Screenshot_20230228_151013

Steps to reproduce:
- display any entity with the "tabs" plugin in sidebar configuration where the child entities does not have a "name" attribute

Expected behavior:
- "tabs" should use the attribute name from the parent as fallback if the child entity does not have a label or name
  - `entity.label ?? beautify(entity.name) ?? attribute.name`

soofstad:
- "tabs" plugin label text now has fallback values if the entity does not have a "name"-attribute

see #101 

closes #101

### DIFF
--- a/packages/dm-core-plugins/src/tabs/Tabs.tsx
+++ b/packages/dm-core-plugins/src/tabs/Tabs.tsx
@@ -4,6 +4,7 @@ import { home } from '@equinor/eds-icons'
 import styled from 'styled-components'
 import { TChildTab } from './TabsContainer'
 import { useTabContext } from './TabsContext'
+import { prettifyName } from './utils'
 
 interface ITabs {
   active: boolean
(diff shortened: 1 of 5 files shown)


packages/dm-core-plugins/src/tabs/Tabs.tsx
```diff
@@ -62,7 +65,9 @@ export const Tabs = (): JSX.Element => {
             onClick={() => setSelectedTab(tabData.attribute)}
             active={selectedTab === tabData.attribute}
           >
-            {tabData.entity.name}
+            {tabData.entity?.label ??
+              prettifyName(tabData.entity?.name || '') ??
+              selectedTab}
```
ingeridhellen, line 70: shouldn't it be 
```suggestion
              tabData.entity?.attribute}
```
?

KristianKjerstad (approved):
looks good! assuming other requested changes are implemented

eoaksnes (approved):

packages/dm-core-plugins/src/tabs/Tabs.tsx
soofstad: It should!

ingeridhellen (approved):

PR #105, opened 2023-03-01
fix: add protocol to type returned from JobHandlerPicker

### LINKED ISSUE #106: fix: bug in JobHandlerPicker

KristianKjerstad:
fix bug in JobHandlerPicker.

JobHandlerPicker is used in Analysis platform
image

the blueprintResolve endpoint in dmss return only a path
image

 the protocol should be added since JobHandlerPicker returns a type
bug
make analysis platform work again
closes #106

### DIFF
--- a/packages/dm-core/src/components/Pickers/JobHandlerPicker.tsx
+++ b/packages/dm-core/src/components/Pickers/JobHandlerPicker.tsx
@@ -12,6 +12,7 @@ export const JobHandlerPicker = (props: {
   const blueprintName = formData.split('/').pop()
   const { token } = useContext(AuthContext)
   const dmssApi = new DmssAPI(token)
+  const protocolPrefix = 'dmss://'
   const [searchResult] = useSearch<any>(
     {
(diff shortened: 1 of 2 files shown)


packages/dm-core/src/components/Pickers/JobHandlerPicker.tsx
```diff
@@ -26,7 +27,7 @@ export const JobHandlerPicker = (props: {
         absoluteId: `WorkflowDS/${blueprintId}`,
       })
       .then((response: any) => {
-        onChange(response.data)
+        onChange(`${protocolPrefix}${response.data}`)
```
soofstad, line 30: Should we not rather fix DMSS to return a path with the correct protocol prefix?

soofstad (changes requested):

packages/dm-core/src/components/Pickers/JobHandlerPicker.tsx
KristianKjerstad: mm, yeah. That makes sense.

PR #120, opened 2023-03-06
refactor: remove typescript errors

eoaksnes:

* Upgrade EDS

* Use same EDS as in dm-core-plugins


soofstad (approved):
Good fix! But should probably use correct event type instead of that custom type

ingeridhellen (approved):

PR #125, opened 2023-03-07
feat: add option to hide attributes in attribute selector plugin

### LINKED ISSUE #119: feat: add to sidebar support to specify attributes
right now the Sidebar component inside AttributeSelector plugin will display all of the entity's attributes in the sidebar. 

Update the ui recipe config such that it can take a list of string that specify what attributes to display in the sidebar.

KristianKjerstad:
add option to hide attributes in the attribute selector plugin.  This can be controlled by the config (see packages/dm-core-plugins/blueprints/tabs/AttributeSelectorPluginConfig.json)
closes #119

### DIFF
--- a/example/app/data/DemoDataSource/DemoPackage/blueprints/Plugins/AttributeSelectorPluginConfig.json
+++ b/example/app/data/DemoDataSource/DemoPackage/blueprints/Plugins/AttributeSelectorPluginConfig.json
@@ -7,7 +7,7 @@
       "type": "CORE:BlueprintAttribute",
       "name": "childTabsOnRender",
       "optional": true,
-      "default": false
+      "default": true
     },
     {
       "attributeType": "string",
@@ -21,7 +21,8 @@
       "type": "CORE:BlueprintAttribute",
       "name": "visibleAttributes",
       "optional": true,
-      "dimensions": "*"
+      "dimensions": "*",
+      "default": []
     },
     {
       "attributeType": "boolean",
--- a/example/app/data/DemoDataSource/DemoPackage/blueprints/Plugins/AttributeSelectorPluginConfig.json
+++ b/example/app/data/DemoDataSource/DemoPackage/blueprints/Plugins/AttributeSelectorPluginConfig.json
@@ -17,6 +17,13 @@
       "default": "home"
     },
     {
+      "attributeType": "string",
+      "type": "CORE:BlueprintAttribute",
+      "name": "visibleAttributes",
+      "optional": true,
+      "dimensions": "*"
+    },
+    {
       "attributeType": "boolean",
       "type": "CORE:BlueprintAttribute",
       "name": "asSidebar",
--- a/packages/dm-core-plugins/src/attribute-selector/AttributeSelectorPlugin.tsx
+++ b/packages/dm-core-plugins/src/attribute-selector/AttributeSelectorPlugin.tsx
@@ -48,12 +48,11 @@ export const AttributeSelectorPlugin = (props: IUIPlugin): JSX.Element => {
     if (config.childTabsOnRender) {
       const newChildTabs: TStringMap = {}
       Object.entries(entity).forEach(([key, attributeData]: [string, any]) => {
-        if (
-          (config?.visibleAttributes === undefined ||
-            config?.visibleAttributes.length === 0 ||
-            config?.visibleAttributes.includes(key)) &&
-          typeof attributeData == 'object'
-        ) {
+        const filteredOutInConfig =
+          config?.visibleAttributes !== undefined &&
+          config?.visibleAttributes.length > 0 &&
+          !config?.visibleAttributes.includes(key)
+        if (!filteredOutInConfig && typeof attributeData == 'object') {
           newChildTabs[key] = {
             attribute: key,
             entity: attributeData,
--- a/packages/dm-core-plugins/src/attribute-selector/AttributeSelectorPlugin.tsx
+++ b/packages/dm-core-plugins/src/attribute-selector/AttributeSelectorPlugin.tsx
@@ -32,7 +32,7 @@ export type TAttributeSelectorPluginConfig = {
 export const AttributeSelectorPlugin = (props: IUIPlugin): JSX.Element => {
   const { idReference, config: passedConfig, onSubmit } = props
   const config: TAttributeSelectorPluginConfig = {
-    childTabsOnRender: passedConfig?.childTabsOnRender ?? false,
+    childTabsOnRender: passedConfig?.childTabsOnRender ?? true,
     homeRecipe: passedConfig?.homeRecipe ?? 'home',
     asSidebar: passedConfig?.asSidebar ?? false,
     visibleAttributes: passedConfig?.visibleAttributes ?? [],
@@ -48,7 +48,12 @@ export const AttributeSelectorPlugin = (props: IUIPlugin): JSX.Element => {
     if (config.childTabsOnRender) {
       const newChildTabs: TStringMap = {}
       Object.entries(entity).forEach(([key, attributeData]: [string, any]) => {
-        if (typeof attributeData == 'object') {
+        if (
+          (config?.visibleAttributes === undefined ||
+            config?.visibleAttributes.length === 0 ||
+            config?.visibleAttributes.includes(key)) &&
+          typeof attributeData == 'object'
+        ) {
           newChildTabs[key] = {
             attribute: key,
             entity: attributeData,
--- a/packages/dm-core-plugins/src/attribute-selector/AttributeSelectorPlugin.tsx
+++ b/packages/dm-core-plugins/src/attribute-selector/AttributeSelectorPlugin.tsx
@@ -26,6 +26,7 @@ export type TAttributeSelectorPluginConfig = {
   childTabsOnRender?: boolean
   homeRecipe?: string
   asSidebar?: boolean
+  visibleAttributes?: string[]
 }
 
 export const AttributeSelectorPlugin = (props: IUIPlugin): JSX.Element => {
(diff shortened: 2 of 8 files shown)


packages/dm-core-plugins/src/attribute-selector/README.md
```diff
@@ -5,4 +5,13 @@ in a new tab.
 
 If you want to specify a config object in some UiRecipe-entity, the Blueprint is included in the npm package.
 
+
+The attribute-selector UI plugin can accept a config of type `AttributeSelectorPluginConfig` with the following attributes:
+
+* `childTabsOnRender` (optional): ??
```
KristianKjerstad, line 11: Could not figure out what childTabsOnRender is used for...

packages/dm-core-plugins/src/attribute-selector/Sidebar.tsx
```diff
@@ -18,19 +25,27 @@ export const Sidebar = (): JSX.Element => {
...
-            onClick={() => setSelectedTab(tabData.attribute)}
-            active={selectedTab === tabData.attribute}
-          />
-        ))}
+        {Object.values(childTabs).map((tabData: TChildTab) => {
+          if (
+            config?.visibleAttributes === undefined ||
+            config?.visibleAttributes.length === 0 ||
+            config?.visibleAttributes.includes(tabData.attribute)
+          ) {
```
KristianKjerstad, line 33: the visibleAttributes list, specified in the config, must include the attribute name for the sidebar entry to render.

The sidebar entry will also render if the visibleAttributes is an empty list of undefined

packages/dm-core-plugins/src/attribute-selector/README.md
ingeridhellen: I think that if `childTabsOnRender == false`, then the tab/sidebar will only show the home button and not the children. I suppose we don't need it anymore, now that we got the `visibleAttributes`? We could make `visibleAttributes` so that if `visibleAttributes == undefined` we show all attributes, but if `visibleAttributes == []`we show none of them.

packages/dm-core-plugins/src/attribute-selector/Sidebar.tsx
ingeridhellen: I wouldn't have the logic here. childTabs is generated in AttributeSelectorPlugin, so I would do a filtering there instead

ingeridhellen (changes requested):

example/app/data/DemoDataSource/DemoPackage/blueprints/Plugins/AttributeSelectorPluginConfig.json
```diff
@@ -16,6 +16,13 @@
       "optional": true,
       "default": "home"
     },
+    {
+      "attributeType": "string",
+      "type": "CORE:BlueprintAttribute",
+      "name": "visibleAttributes",
+      "optional": true,
+      "dimensions": "*"
```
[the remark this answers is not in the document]
soofstad: There is no hard rule, but yes, we should.
The code that actually reads the "default" value is not very mature, but it's still a good idea to add it

packages/dm-core-plugins/src/attribute-selector/README.md
soofstad: I think "null | undefined| []" should be treated the same. Showing all attributes.
Since the default value for "visibleAttributes" is "[]", that should result in showing all.
Goal being that "stuff just works" when there is no special config and default values.

packages/dm-core-plugins/src/attribute-selector/AttributeSelectorPlugin.tsx
```diff
@@ -34,6 +35,7 @@ export const AttributeSelectorPlugin = (props: IUIPlugin): JSX.Element => {
     childTabsOnRender: passedConfig?.childTabsOnRender ?? false,
     homeRecipe: passedConfig?.homeRecipe ?? 'home',
     asSidebar: passedConfig?.asSidebar ?? false,
+    visibleAttributes: passedConfig?.visibleAttributes ?? [],
```
soofstad, line 38: think we should change the "childTabsOnRender" to default to "true" now.

soofstad (changes requested):

packages/dm-core-plugins/src/attribute-selector/README.md
ingeridhellen: Can't we just change the default value to null, then?

soofstad: I dont think we should remove "childTabsOnRender". Could still be useful for entities with a ton of attributes.
One problem is that default values for lists will always be "[]", as instantiated from DMSS.

KristianKjerstad:
@soofstad @ingeridhellen feel free to take over this PR since I will not be able to fix review suggestions this week :)

example/app/data/DemoDataSource/DemoPackage/blueprints/Plugins/AttributeSelectorPluginConfig.json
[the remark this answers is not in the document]
soofstad: On second thought; no. All basic types has a "default default" value. For strings it's "".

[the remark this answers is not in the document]
ingeridhellen: Isn't explicit always better than implicit?

packages/dm-core-plugins/src/attribute-selector/AttributeSelectorPlugin.tsx
```diff
@@ -26,14 +26,16 @@ export type TAttributeSelectorPluginConfig = {
...
   homeRecipe?: string
   asSidebar?: boolean
+  visibleAttributes?: string[]
 }
 
 export const AttributeSelectorPlugin = (props: IUIPlugin): JSX.Element => {
   const { idReference, config: passedConfig, onSubmit } = props
   const config: TAttributeSelectorPluginConfig = {
-    childTabsOnRender: passedConfig?.childTabsOnRender ?? false,
+    childTabsOnRender: passedConfig?.childTabsOnRender ?? true,
```
ingeridhellen, line 35: I agree with the change, but I think you should also update the default value in the AttributeSelectorPluginConfig files

packages/dm-core-plugins/src/attribute-selector/AttributeSelectorPlugin.tsx
```diff
@@ -46,7 +48,12 @@ export const AttributeSelectorPlugin = (props: IUIPlugin): JSX.Element => {
     if (config.childTabsOnRender) {
       const newChildTabs: TStringMap = {}
       Object.entries(entity).forEach(([key, attributeData]: [string, any]) => {
-        if (typeof attributeData == 'object') {
+        if (
+          (config?.visibleAttributes === undefined ||
+            config?.visibleAttributes.length === 0 ||
+            config?.visibleAttributes.includes(key)) &&
+          typeof attributeData == 'object'
+        ) {
```
ingeridhellen, line 56: Perhaps more readable?
```suggestion
        const filteredOutInConfig = 
          config?.visibleAttributes !== undefined &&
          config?.visibleAttributes.length > 0 &&
          !config?.visibleAttributes.includes(key)
        if !filteredOutInConfig && typeof attributeData == 'object') {
```

packages/dm-core-plugins/src/attribute-selector/README.md
ingeridhellen:
```suggestion
* `childTabsOnRender` (optional): If false the tab/sidebar will only show the home button and not the attributes. Defaults to true.
```

packages/dm-core-plugins/src/attribute-selector/Sidebar.tsx
```diff
@@ -18,19 +19,21 @@ export const Sidebar = (): JSX.Element => {
...
-              tabData.entity?.label ||
-              prettifyName(tabData.entity?.name || '') ||
-              tabData.attribute
-            }
-            onClick={() => setSelectedTab(tabData.attribute)}
-            active={selectedTab === tabData.attribute}
-          />
-        ))}
+        {Object.values(childTabs).map((tabData: TChildTab) => {
+          return (
```
ingeridhellen, line 23: Could remove the extra curly brackets now

packages/dm-core-plugins/src/attribute-selector/Tabs.tsx
```diff
@@ -56,23 +62,31 @@ export const Tabs = (): JSX.Element => {
...
-          </ChildTab>
-        </Tooltip>
-      ))}
+      {Object.values(childTabs).map((tabData: TChildTab) => {
+        if (
+          config?.visibleAttributes === undefined ||
+          config?.visibleAttributes.length === 0 ||
+          config?.visibleAttributes.includes(tabData.attribute)
+        ) {
+          return (
```
ingeridhellen, line 71: This is already filtered on AttributeSelectorPlugin

ingeridhellen (changes requested):

packages/dm-core-plugins/src/attribute-selector/AttributeSelectorPlugin.tsx
soofstad: fixed

packages/dm-core-plugins/src/attribute-selector/AttributeSelectorPlugin.tsx
soofstad: fixed

packages/dm-core-plugins/src/attribute-selector/Sidebar.tsx
soofstad: fixed

packages/dm-core-plugins/src/attribute-selector/Tabs.tsx
soofstad: fixed

packages/dm-core-plugins/src/attribute-selector/AttributeSelectorPlugin.tsx
ingeridhellen: You forgot the one in the example folder (uppermost file in this PR)

ingeridhellen (changes requested):
Looks good! One remaining though

ingeridhellen (approved):

PR #130, opened 2023-03-09
feat: recipe hook

eoaksnes:

* Add useRecipe hook

* Get recipe using re-usable hook


packages/dm-core/src/hooks/useRecipe.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * Hook to get a recipe
+ *
+ * @docs Hooks
+ *
+ * @usage
+ * Code example:
+ * ```
+ * import { useRecipe } from '@data-modelling-tool/core'
```
ingeridhellen, line 42:
```suggestion
 * import { useRecipe } from '@development-framework/dm-core'
```

packages/dm-core/src/hooks/useRecipe.tsx
```diff
@@ -0,0 +1,68 @@
...
+ * }
+ * ```
+ *
+ * @param typeRef - The reference to the blueprint to retrieve a recipe for
+ * @param recipeName - Name of recipe to find (optional)
+ * @returns A list containing the blueprint document, a boolean representing the loading state, and an Error, if any.
+ */
+export const useRecipe = (
+  typeRef: string,
+  recipeName: string | null
```
ingeridhellen, line 59: This way, you would have to send in `recipeName == null ` if you would want to use the default one. Write instead 
```suggestion
  recipeName?: string
```
, then you can skip the input variable if yoou don't want to specify

packages/dm-core/src/hooks/useRecipe.tsx
```diff
@@ -0,0 +1,68 @@
+import React from 'react'
+import { useBlueprint, TUiRecipe } from '../index'
+
+const findRecipe = (
+  initialUiRecipe: any,
+  recipeName: string | null,
+  recipes: any[]
+) => {
```
ingeridhellen, line 8:
```suggestion
  initialUiRecipe: TUiRecipe | undefined,
  recipeName: string | undefined,
  recipes: TUiRecipe[]
): TUiRecipe => {
```

soofstad (changes requested):

packages/dm-core/src/hooks/useRecipe.tsx
```diff
@@ -0,0 +1,68 @@
...
+    return initialUiRecipe
+  }
+  if (recipes.length > 0) {
+    return recipes[0]
+  }
+  return {
+    name: 'yaml',
+    plugin: 'yaml',
+    config: {},
+  }
```
ingeridhellen, line 25:
```suggestion
  return {
    type: 'CORE:UIRecipe',
    name: 'yaml',
    plugin: 'yaml',
    config: {},
  }
```

ingeridhellen (changes requested):
Looks good! Some minor adjustments on typing and comments, though

packages/dm-core/src/hooks/useRecipe.tsx
```diff
@@ -0,0 +1,68 @@
...
+  recipeName: string | null,
+  recipes: any[]
+) => {
+  // If recipe is defined, find and return the ui recipe from available recipes.
+  if (recipeName) {
+    return recipes.find((recipe: any) => recipe.name == recipeName)
+  }
+  // If not recipe is defined, use initialize recipe, or the first from recipes list or lastly fallback.
+  // TODO: how to check that is empty?
+  if (initialUiRecipe) {
```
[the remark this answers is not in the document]
eoaksnes: Good catch, I have added a check that the initialUiRecipe keys should have a length > 0.

ingeridhellen (approved):

PR #131, opened 2023-03-09
feat: add grid plugin

eoaksnes:

* Add grid plugin

* Create dashboards or place items using grid plugin


packages/dm-core-plugins/src/grid/GridItems.tsx
```diff
@@ -0,0 +1,25 @@
...
+import React from 'react'
+import { GridElement } from './GridElement'
+
+type GridItemsProps = {
+  items: TGridItem[]
+  idReference: string
+  document: TGenericObject
+}
+
+export const GridItems = (props: GridItemsProps) => {
```
soofstad, line 12: A bit overkill having this as a separate file/class?

soofstad (approved):

PR #132, opened 2023-03-09
feat: view creator

eoaksnes:

* Create views from view configs

* Create a view config to specify what and how to view


example/app/data/DemoDataSource/DemoPackage/blueprints/Plugins/View/RecipeViewConfig.json
```diff
@@ -0,0 +1,12 @@
+{
+  "type": "dmss://system/SIMOS/Blueprint",
+  "name": "RecipeViewConfig",
```
soofstad, line 3: Is "InlineRecipeViewConfig" or "ProvidedRecipeViewConfig" a better word? :thinking:

packages/dm-core/src/components/ViewCreator/ViewCreator.tsx
```diff
@@ -0,0 +1,70 @@
...
+} from '../../types'
+import { TGenericObject, UiPluginContext } from '../../index'
+import React, { useContext } from 'react'
+import { ReferenceView } from './ReferenceView'
+import { RecipeView } from './RecipeView'
+import { getTarget, getType } from './utils'
+
+type TViewCreator = {
+  idReference: string
+  document: TGenericObject
```
soofstad, line 14: Should it be up to the parent, or ViewCreator to fetch the document, at the correct depth.
I think it should be handled by "ViewCreator", to make the component as simple as possible to use.

packages/dm-core/src/components/ViewCreator/ViewCreator.tsx
```diff
@@ -0,0 +1,70 @@
...
+  viewConfig: TViewConfig
+}
+
+/**
+ * A component that will create a view from a view config.
+ *
+ * A view config can be a recipe (RecipeViewConfig) or reference to existing recipe (ReferenceViewConfig).
```
soofstad, line 21:
```suggestion
 * A view config can contain a UIRecipe (RecipeViewConfig) or reference an existing one (ReferenceViewConfig).
```

packages/dm-core/src/types.ts
```diff
@@ -92,3 +92,30 @@ export type TPlugin = {
...
+
+export type TReferenceViewConfig = TViewConfig & {
+  recipe: string
+}
+
+export type TRecipeViewConfig = TViewConfig & {
+  recipe: TUiRecipe
+}
+
+export function isReferenceViewConfig(
```
soofstad, line 111: Don't you only need one of these type predicates?
If its not one, it's always the other

soofstad (changes requested):

example/app/data/DemoDataSource/DemoPackage/blueprints/Plugins/View/RecipeViewConfig.json
eoaksnes: I liked this, it is much more precise. 

I renamed it to InlineRecipeViewConfig, but I liked the other alternative also.

packages/dm-core/src/components/ViewCreator/ViewCreator.tsx
eoaksnes: I also think this should be up to the ViewCreator. Right now the ViewCreator only fetches the document to get the list of UI recipes. 
Not sure if we need to fetch it for something else, maybe when we will support dotted paths for scope?

eoaksnes: I guess it is possible to add depth as an attribute in the ViewConfig blueprint so that it can be used as input to either control the depth inside ViewCreator or passed to the plugin to control the depth for what happens inside the plugin.

soofstad (approved):

packages/dm-core/src/components/ViewCreator/ViewCreator.tsx
eoaksnes: Added issue for this: https://github.com/equinor/dm-core-packages/issues/133

PR #149, opened 2023-03-14
fix: content of sidebar now fills available space

### LINKED ISSUE #148: fix(attribute-sidebar): width of content does not use all available space

soofstad:
- set CSS prop "fill-available" on some wrappers

- some plugins were being squashed

Before:
Screenshot_20230314_091639

After:
Screenshot_20230314_091703

closes #148

### DIFF
--- a/packages/dm-core-plugins/src/attribute-selector/AttributeSelectorPlugin.tsx
+++ b/packages/dm-core-plugins/src/attribute-selector/AttributeSelectorPlugin.tsx
@@ -90,6 +90,7 @@ export const AttributeSelectorPlugin = (props: IUIPlugin): JSX.Element => {
         style={{
           display: 'flex',
           flexDirection: config.asSidebar ? 'row' : 'column',
+          width: '100%',
         }}
       >
         {config.asSidebar ? <Sidebar /> : <Tabs />}
--- a/packages/dm-core-plugins/src/attribute-selector/Content.tsx
+++ b/packages/dm-core-plugins/src/attribute-selector/Content.tsx
@@ -11,6 +11,7 @@ import { TChildTab } from './AttributeSelectorPlugin'
 const HidableWrapper = styled.div<any>`
   display: ${(props: { hidden: boolean }) => (props.hidden && 'none') || 'flex'}
   align-self: normal;
+  width: 100%;
 `
 
 export const Content = (): JSX.Element => {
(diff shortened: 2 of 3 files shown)


packages/dm-core-plugins/src/attribute-selector/AttributeSelectorPlugin.tsx
```diff
@@ -90,6 +90,7 @@ export const AttributeSelectorPlugin = (props: IUIPlugin): JSX.Element => {
         style={{
           display: 'flex',
           flexDirection: config.asSidebar ? 'row' : 'column',
+          width: '-webkit-fill-available',
```
ingeridhellen, line 93:
```suggestion
```

packages/dm-core-plugins/src/attribute-selector/Content.tsx
```diff
@@ -11,6 +11,7 @@ import { TChildTab } from './AttributeSelectorPlugin'
 const HidableWrapper = styled.div<any>`
   display: ${(props: { hidden: boolean }) => (props.hidden && 'none') || 'flex'}
   align-self: normal;
+  width: -webkit-fill-available;
```
ingeridhellen, line 14:
```suggestion
  width: 100%;
```

packages/dm-core-plugins/src/attribute-selector/Content.tsx
```diff
@@ -26,7 +27,7 @@ export const Content = (): JSX.Element => {
     config,
   } = useTabContext()
   return (
-    <>
+    <div style={{ width: '-webkit-fill-available' }}>
```
ingeridhellen, line 30:
```suggestion
    <div style={{ width: '100%' }}>
```

ingeridhellen (changes requested):
Have you tested multiple browsers? I believe -webkit-fill-available only works on some browsers. I would instead like you to do the below fixes

packages/dm-core-plugins/src/attribute-selector/AttributeSelectorPlugin.tsx
soofstad: This one is needed. If not, the "fill avilable" only effects the first level of the sidebar (if they are nested).
But "100%" works on chrome and firefox :+1:

ingeridhellen (approved):

PR #150, opened 2023-03-15
fix(form): fix a bug where form continued with bad state

soofstad:
- "form" will throw error on failed blueprint fetches
- update ErrorBoundary to display error and plugin name

Screenshot_20230317_121053

- "form" would silently continue even if it failed to fetch a blueprint. Making it hard to detect errors/bad config

none

### DIFF
--- a/packages/dm-core-plugins/package.json
+++ b/packages/dm-core-plugins/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@development-framework/dm-core-plugins",
   "license": "MIT",
-  "version": "1.0.5",
+  "version": "1.0.6",
   "main": "dist/index.js",
   "dependencies": {
     "@development-framework/dm-core": "^1.0.41",
--- a/packages/dm-core-plugins/src/form/FormPlugin.tsx
+++ b/packages/dm-core-plugins/src/form/FormPlugin.tsx
@@ -21,9 +21,10 @@ const ErrorHelperText = styled.div`
 const widgets = {
   TypeWidget: (props: TWidget) => {
     const { id, namePath, label, value } = props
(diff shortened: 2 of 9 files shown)


packages/dm-core-plugins/src/form/components/ErrorWrapper.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+  return (
+    <ErrorGroup>
+      <b>{header}</b>
+      <p>Failed to render form</p>
+      <button
+        style={{ width: '150px' }}
+        onClick={() => setIsDetailsHidden(!isDetailsHidden)}
+      >
+        Show details
```
ingeridhellen, line 25: Why not use the EDS button?

ingeridhellen (approved):
I wonder though: Normally when making a web app, we wrap the entire app in an ErrorBoundary which catches all errors and displays an error page. Could we do that here? We could upgrade create-dm-app to include an ErrorBoundary. Then the form plugin could just raise an error if useBlueprint returns an error. Would mean the entire page would be covered by the error message though, not just the relevant field.Could solve that by wrapping individual components in an error boundary, though

soofstad:
> I wonder though: Normally when making a web app, we wrap the entire app in an ErrorBoundary which catches all errors and displays an error page. Could we do that here? We could upgrade create-dm-app to include an ErrorBoundary. Then the form plugin could just raise an error if useBlueprint returns an error. Would mean the entire page would be covered by the error message though, not just the relevant field.Could solve that by wrapping individual components in an error boundary, though

Completely agree. We already have a ErrorBoundary. Not sure why I did not think of using that...
Will rewrite.

PR #164, opened 2023-03-21
fix: make tree view and right window in explorer plugin scrollable

### LINKED ISSUE #128: bug(explorer): right plugin window is not vertically scrollable

KristianKjerstad:
make it possible to scroll in left and right side panels in the explorer plugin.

closes #128

### DIFF
--- a/packages/dm-core-plugins/src/explorer/style.css
+++ b/packages/dm-core-plugins/src/explorer/style.css
@@ -1,3 +1,3 @@
 body .lm_content {
-  overflow: scroll;
+  overflow: auto;
 }
--- a/packages/dm-core-plugins/src/explorer/style.css
+++ b/packages/dm-core-plugins/src/explorer/style.css
@@ -1,0 +1,3 @@
+body .lm_content {
+  overflow: scroll;
+}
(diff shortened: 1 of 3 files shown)


packages/dm-core-plugins/src/explorer/style.css
```diff
@@ -0,0 +1,3 @@
+body .lm_content {
```
KristianKjerstad, line 1: this will make golden layout panels scrollable

example in https://github.com/golden-layout/golden-layout/issues/29

packages/dm-core-plugins/src/explorer/style.css
```diff
@@ -0,0 +1,3 @@
...
+  overflow: scroll;
```
soofstad, line 2: Same goes for this one right?
I should have pointed out both right away

soofstad (approved):

packages/dm-core-plugins/src/explorer/style.css
KristianKjerstad: ups, forgot to push that change. Will fix!

Issue #151, opened 2023-03-15
Feature (views): Add concept of views to attribute-selector and recipe-selector


l3abak:
#### Suggest Solution
The same is Grid plugin, the concept of selecting a plugin for based on different recipes and for different attributes could be solved using the "View" concept. 

Adding that would merge the two plugins and create a "Selector" plugin which would be much easier to use and maintain. 

#### Additional Details
- [x] recipe-selector
- [ ] attribute-selector

eoaksnes:
Add icon after merging the plugins.

PR #165, opened 2023-03-21
feature(UiRecipeSelector): support view concept

soofstad:
- Refactor of UiRecipeSelector to support "view concept". Renamed plugin to "recipe-selector"
- Delete UiPluginSelector"

- Plugins intended to display chidlren (attributes or recipes) should use the "view" concept

relates to #151


packages/dm-core/src/components/RecipeSelector.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+    if (config.views && config.views.length) {
+      setSelectableViews(config.views)
+    } else {
+      // If no views are passed. Create from recipes
+      const newSelectableViews: (
+        | TReferenceViewConfig
+        | TInlineRecipeViewConfig
+      )[] = []
+      uiRecipes.forEach((recipe: TUiRecipe) => {
```
eoaksnes, line 71: is it better to use .map? 

const newSelectableViews = uiRecipes.map(...)

eoaksnes (approved):
Nice, just one comment

packages/dm-core/src/components/RecipeSelector.tsx
soofstad: Cant, as we are doing a check before pushing to the array.

PR #170, opened 2023-03-23
Feature/view concept for attribute selector

### LINKED ISSUE #151: Feature (views): Add concept of views to attribute-selector and recipe-selector

soofstad:
- Total rewrite of "attribute-selector" - Can now be configured by "ViewConfigs"
- Fix bug in "form" where the wrapper did not use 100% width
- Fix a bug in "form" where it would not check for "error" when calling "useDocument"
 
- A more standardized way of configuring plugins designed to accomodate multiple views 

closes #151 closes #170


packages/dm-core/src/types.ts
```diff
@@ -98,6 +98,7 @@ export type TUserIdMapping = { userId: string; username: string }
 export type TViewConfig = {
   type: string
   scope: string
+  label?: string
```
eoaksnes, line 101: Not sure if we should add a label attribute to ViewConfig, my impression is that these things are plugin-specific. 

So maybe extends the TViewConfig or create a new type that includes the view config? Then this will be plugin specific for attribute-selector that can also include a label. 

Extends:
```
"type": "AttributeSelectorItem",
"label":"",
 ...viewConfig...
```

Compose:
```
"type": "AttributeSelectorItem",
"label": "",
"viewConfig": {
  ...
}
```

eoaksnes (changes requested):

packages/dm-core/src/types.ts
soofstad: fixed

eoaksnes (approved):
Nice

PR #171, opened 2023-03-23
feature(ViewCreator): update to handle views with no recipe

soofstad:
- Added option for "useRecipe" to not use "initialRecipes" (needed to avoid recursion on generated fall back ViewConfigs)
- Support for ViewConfigs with no recipe (needed for generated ViewConfigs based on attribute only)
- Partial support for special scope keyword 'self'
 

- A more standardized way of configuring plugins designed to accomodate multiple views 

needed for #151


eoaksnes (approved):
Nice

Issue #109, opened 2023-03-02
feat: show list of children using a ui recipe specified in the child


eoaksnes:
To show a list of objects a new generic UI plugin is needed. 

![image](https://user-images.githubusercontent.com/42837660/225619620-23026961-7a89-4e4a-a9c5-65b8233c71e7.png)

* Collapsible Object Table
- [x] Each object is shown as a row. 
- [x] There is a config element to define what items to show as rows for each child on the banner
- [x] The ui-plugin inside each element is defined based on the "View" concept as defined in the Grid plugin.
- [x] The table has general functions such as;
  - [x]  add
  - [x] delete
  - [x] move (reorder)
- [ ] (optional) Config allows for adding buttons, icon and call-back function, at the end of each row.
- [x] Default behaviour is to get the default view of each component. 
- [ ] Add pagination (view only) @soofstad 

* Object Table
Same as above, without the collapsible part. Instead the rows include editable data fields. 
- [x] There is a config element to define what items to to show on the columns
- [x] config has an option to be editable to read only. 
- [x] The table has general functions such as, add, delete, and move (reorder)
- [ ] (optional) Config allows for adding buttons, icon and call-back function, at the end of each row.
- [ ] Add pagination (view only)
 
Related issues:
- https://github.com/equinor/dm-core-packages/issues/127

eoaksnes:
https://mui.com/material-ui/react-table/#collapsible-table

PR #173, opened 2023-03-28
feature(generic-list): added generic list plugin

soofstad:
- Adds a new plugin "generic-list"
- Update "useRecipe" to select a recipe with a matching "dimensions"

Screenshot_20230328_103456

- Framework should be able to render attributes of entities, even when they are lists

fixes parts of #109

### DIFF
--- a/packages/dm-core-plugins/blueprints/generic-list/GenericListPluginConfig.json
+++ b/packages/dm-core-plugins/blueprints/generic-list/GenericListPluginConfig.json
@@ -1,0 +1,12 @@
+{
+  "type": "CORE:Blueprint",
+  "name": "GenericListPluginConfig",
+  "attributes": [
+    {
+      "attributeType": "boolean",
+      "type": "CORE:BlueprintAttribute",
+      "name": "expanded",
+      "default": false
+    }
+  ]
+}
--- a/packages/dm-core-plugins/src/generic-list/GenericListPlugin.tsx
+++ b/packages/dm-core-plugins/src/generic-list/GenericListPlugin.tsx
@@ -1,0 +1,51 @@
+import React from 'react'
+import {
+  IUIPlugin,
+  Loading,
+  TGenericObject,
+  useDocument,
+  EntityView,
+} from '@development-framework/dm-core'
+
+import { Accordion } from '@equinor/eds-core-react'
+
+type TGenericListConfig = {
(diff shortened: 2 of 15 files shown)


packages/dm-core-plugins/src/generic-list/GenericListPlugin.tsx
```diff
@@ -0,0 +1,84 @@
...
+  const internalConfig = { ...defaultConfig, ...config }
+  const [document, loading, _, error] = useDocument<TGenericObject[]>(
+    idReference,
+    2
+  )
+  if (loading) return <Loading />
+
+  if (error) throw new Error(JSON.stringify(error, null, 2))
+  if (!document) return <Loading />
+  if (internalConfig?.collapsable)
```
ingeridhellen, line 49:
```suggestion
  if (internalConfig.collapsable)
```

packages/dm-core-plugins/blueprints/generic-list/GenericListPluginConfig.json
```diff
@@ -0,0 +1,12 @@
+{
+  "type": "CORE:Blueprint",
+  "name": "GenericListPluginConfig",
+  "attributes": [
+    {
+      "attributeType": "boolean",
+      "type": "CORE:BlueprintAttribute",
+      "name": "collapsable",
```
ingeridhellen, line 8: Are we sure we want to merge Accordion and List into one Plugin? They don't seem very related, except that we use them quite the same way.

soofstad: Agreed. Removed config option. Only get collapsible ones now.

packages/dm-core/src/components/EntityView.tsx
```diff
@@ -12,14 +12,27 @@ const Wrapper = styled.div`
...
 export const EntityView = (props: IEntityView): JSX.Element => {
-  const { idReference, type, onSubmit, onOpen, recipeName, noInit } = props
+  const {
+    idReference,
+    type,
+    onSubmit,
+    onOpen,
+    recipeName,
+    noInit,
+    dimensions,
```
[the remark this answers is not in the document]
soofstad: Added example that needs this in App.js

ingeridhellen (approved):

PR #175, opened 2023-03-30
Feature/generic list plugin2

soofstad:

- Adds "add" and "remove" features to "generic-list"
- Reuse "ErrorGroup" component where applicable
- Have "useRecipe" raise proper error when it fails

- Basic remove and add functionality to "generic-list"

base functionality for #109

### DIFF
--- a/packages/dm-core-plugins/src/generic-list/GenericListPlugin.tsx
+++ b/packages/dm-core-plugins/src/generic-list/GenericListPlugin.tsx
@@ -9,14 +9,40 @@ import {
   useDocument,
 } from '@development-framework/dm-core'
 import { AxiosResponse } from 'axios'
-import { Accordion, Button, Icon } from '@equinor/eds-core-react'
-import { delete_to_trash, library_add } from '@equinor/eds-icons'
+import { Button, Icon } from '@equinor/eds-core-react'
+import {
+  chevron_down,
+  chevron_right,
+  delete_to_trash,
+  library_add,
+} from '@equinor/eds-icons'
+
+import styled from 'styled-components'
+
+const TableData = styled.td`
+  text-align: right;
+`
+const TableRow = styled.tr`
+  text-align: right;
+  border-collapse: collapse;
+`
+const Table = styled.table`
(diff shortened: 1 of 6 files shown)


packages/dm-core-plugins/src/generic-list/GenericListPlugin.tsx
```diff
@@ -25,27 +54,154 @@ export const GenericListPlugin = (
...
+        key,
+        internalConfig.expanded,
+      ])
+    )
+    // const itemsExpanded = itemsWithIds.reduce(
+    //   (array: TGenericObject[], item: TGenericObject) => ({
+    //     ...array,
+    //     [item.__react_key__]: internalConfig.expanded,
+    //   })
+    // )
```
KristianKjerstad, line 91: remove comments?

packages/dm-core-plugins/src/generic-list/GenericListPlugin.tsx
```diff
@@ -25,27 +54,154 @@ export const GenericListPlugin = (
...
+      setItems({ ...items })
+    })
+  }
+  const addItem = (reference: string, type: string) => {
+    dmssAPI
+      // TODO: Get type from parent blueprint, be able to select specialised type
+      .instantiateEntity({
+        entity: { type: items[Object.keys(items)[0]].type },
+      })
+      .then((newEntity: AxiosResponse<any>) => {
```
KristianKjerstad, line 108: according to generated API, typ should be 
` AxiosResponse<object, any>`

packages/dm-core-plugins/src/generic-list/GenericListPlugin.tsx
```diff
@@ -25,27 +54,154 @@ export const GenericListPlugin = (
     idReference,
     2
   )
-  if (loading) return <Loading />
+  const [items, setItems] = useState<TGenericObject[]>([])
```
KristianKjerstad, line 57: is TGenericObject[] type correct for items? 

in the code below it looks like items is a dictionary with a random UID as key + dictionary as value, and not a list.

PR #180, opened 2023-04-14
feature(generic-list): reorder items in list

soofstad:
- Adds "reorder" functionality
- Refactor "generic-list" and "generic-table" to share some similar components

Screenshot_20230414_120446
(note: image is outdated. There is now some padding around the "save"-button)

basic functionality as described in #109 

part of #109

### DIFF
--- a/packages/dm-core-plugins/src/generic-list/Components.tsx
+++ b/packages/dm-core-plugins/src/generic-list/Components.tsx
@@ -1,0 +1,44 @@
+import React, { MouseEvent } from 'react'
+import { Button, Icon, Progress } from '@equinor/eds-core-react'
+import {
+  arrow_down,
+  arrow_up,
+  delete_to_trash,
+  library_add,
+} from '@equinor/eds-icons'
+
+export const AppendButton = (props: {
+  onClick: (event: MouseEvent<HTMLButtonElement>) => void
+}) => (
+  <Button variant="ghost_icon" onClick={props.onClick}>
+    <Icon data={library_add} title="Append" />
+  </Button>
+)
+
+export const SaveButton = (props: {
+  onClick: (event: MouseEvent<HTMLButtonElement>) => void
+  disabled: boolean
+  isLoading: boolean
+}) => (
+  <Button disabled={props.disabled} onClick={props.onClick}>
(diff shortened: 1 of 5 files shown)


packages/dm-core-plugins/src/generic-list/Components.tsx
```diff
@@ -0,0 +1,44 @@
...
+  isLoading: boolean
+}) => (
+  <Button disabled={props.disabled} onClick={props.onClick}>
+    {props.isLoading ? <Progress.Dots color={'primary'} /> : 'Save'}
+  </Button>
+)
+
+export const DeleteButton = (props: { onClick: () => void }) => (
+  <Button variant="ghost_icon" color="danger" onClick={props.onClick}>
+    <Icon data={delete_to_trash} title="Delete" />
```
KristianKjerstad, line 30: not sure if the delete action takes a long time, but it might be nice to add a loading icon here too, same as in SaveButton

packages/dm-core-plugins/src/generic-list/GenericListPlugin.tsx
```diff
@@ -64,6 +67,8 @@ export const GenericListPlugin = (
   const [itemsExpanded, setItemsExpanded] = useState<{
     [key: string]: boolean
   }>({})
+  const [dirtyOrder, setDirtyOrder] = useState<boolean>(false)
```
KristianKjerstad, line 70: I was a bit confused about the `dirtyOrder` variable name

Maybe something like "areItemsReordered" is better.

packages/dm-core-plugins/src/generic-list/Components.tsx
```diff
@@ -0,0 +1,44 @@
+import React, { MouseEvent } from 'react'
```
KristianKjerstad, line 1: Good choice to split the buttons out to a new file!

KristianKjerstad (approved):
Overall good! Just a few remarks you can consider

packages/dm-core-plugins/src/generic-list/GenericListPlugin.tsx
soofstad: renamed to "unsavedChanges"

PR #183, opened 2023-04-19
fix(attribute-selector): pass "onOpen" to views

### LINKED ISSUE #127: Form plugin wrapped in tabs needs improvement for arrays
#### Describe Problem
- Arrays appear both inside the home view, and as tabs for children. 
- not possible to close tabs, for children or array items.
- Cannot open children 
#### Suggest Solution
By default, arrays should only be shown as new tabs, and not as part of the home view. They are to be considered complex objects, even if the type of the array is simple. 
By default, the "form" view of the home should contain all simple attributed for editing, and provide actions, 
"open" for all required complex attributes, 
"remove" and "create" for all optional attributes, simple or complex. 

The "form" plugin needs additional options.
- List of children to be shown, and their related views. The default behaviour include all children, as described above. But this can be modified by specifically.
- option to open children in tabs, and provide a close bottom or not. Default is dictated by children type, and general option for opening them in tabs. 
- option to provide a desired list of actions to be added for complex children.  Default as described above.
- Improve the Array view to include two modes,
- - Giving list of attributes to view in a table. 
- - -Possibility to select and add from a list of "Actions"
- - Show each item in a separate drop-down segment, where the view could be selected using a UIRecipe.
-   
#### Additional Details

soofstad:
- Update "attribute-selector" plugin to pass "onOpen" to Views
- Update "form"-plugins ArrayField to not render lists of complex lists if "onOpen" is passed

- Changes to "view-concept" and "attribute-selector" caused "onOpen" to no longer function

closes #127

### DIFF
--- a/packages/dm-core-plugins/src/attribute-selector/AttributeSelectorPlugin.tsx
+++ b/packages/dm-core-plugins/src/attribute-selector/AttributeSelectorPlugin.tsx
@@ -19,7 +19,7 @@ import {
 export const AttributeSelectorPlugin = (
   props: IUIPlugin & { config?: TAttributeSelectorConfig }
 ): JSX.Element => {
-  const { idReference, config, onSubmit } = props
+  const { idReference, config } = props
   const internalConfig: TAttributeSelectorConfig = {
     childTabsOnRender: true,
     asSidebar: false,
@@ -27,9 +27,24 @@ export const AttributeSelectorPlugin = (
     ...config,
   }
   const [entity, isLoading, _, error] = useDocument<TGenericObject>(idReference)
-  const [selectedView, setSelectedView] = useState<number>(0)
+  const [selectedView, setSelectedView] = useState<string | undefined>()
   const [views, setViews] = useState<TItemData[]>([])
   const [formData, setFormData] = useState<TGenericObject>({})
+
+  function addView(viewId: string, view: TViewConfig) {
+    if (!views.find((view: TItemData) => view.viewId === viewId)) {
+      // View does not exist, add it
+      const newView: TItemData = {
+        viewId: viewId,
+        view: view,
+        label: view.label ?? viewId,
+        rootEntityId: idReference,
+        onSubmit: () => undefined,
+      }
+      setViews([...views, newView])
+    }
+    setSelectedView(viewId)
+  }
 
   useEffect(() => {
     if (!entity) return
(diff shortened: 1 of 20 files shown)


packages/dm-core-plugins/src/attribute-selector/AttributeSelectorPlugin.tsx
```diff
@@ -19,89 +19,89 @@ import {
...
+  const { idReference, config } = props
   const internalConfig: TAttributeSelectorConfig = {
     childTabsOnRender: true,
     asSidebar: false,
     items: [],
     ...config,
   }
   const [entity, isLoading, _, error] = useDocument<TGenericObject>(idReference)
-  const [selectedView, setSelectedView] = useState<number>(0)
+  const [selectedView, setSelectedView] = useState<string>()
```
KristianKjerstad, line 30: should this be initialized to empty string? Otherwise I think type is `string | undefined`

packages/dm-core-plugins/src/form/fields/ArrayField.tsx
```diff
@@ -74,87 +68,41 @@ export default function Fields(props: any) {
...
-        <Button
-          variant="outlined"
-          data-testid={`add-${namePath}`}
-          onClick={() => {
-            handleAddObject()
-          }}
-        >
-          Add
-        </Button>
+        <OpenObjectButton namePath={`${namePath}`} />
```
KristianKjerstad, line 74: props to this component is of type any, but I would assume the type for namePath is a string (?)

Then it would be equivalent to write
```
<OpenObjectButton namePath={namePath} />
```

KristianKjerstad (approved):
overall good!

Just 2 comments

packages/dm-core-plugins/src/attribute-selector/types.tsx
```diff
@@ -10,7 +10,10 @@ export type TAttributeSelectorItem = {
   label?: string
 }
 export type TItemData = TAttributeSelectorItem & {
+  key: string
```
eoaksnes, line 13: The`key` is a generic name, should it be renamed to viewId, itemId (or something like that)? Or, could the key be generated from the information that exists in the view config? Perhaps combine the name of the view config and the scope to get the unique key?

eoaksnes (approved):
Nice! Some comments to consider.

packages/dm-core-plugins/src/attribute-selector/AttributeSelectorPlugin.tsx
soofstad: That is implicit typing for the "useState" with no initial value. But added explicitly now as well

Issue #174, opened 2023-03-29
refactor: create "new DmssAPI(token)" context


soofstad:
Should avoid having to instantiate many DmssAPI's like;
```js
  const { token } = useContext(AuthContext)
  const dmssAPI = new DmssAPI(token)
```

Rather just be able to do:
```js
const dmssAPI = useDMSS() //useDMSS is a context, not a hook

```

eoaksnes:
```
export function useDMSS() {
const { token } = useContext(AuthContext)
const todoAPI = useMemo(() => new TodoAPI(token), [token])
...
}
```

PR #184, opened 2023-04-20
refactor: dmsscontext

ingeridhellen:

Creates a DMSSContext
Replace old `new DmssAPI(token)` syntax with the new context

It's unfortunate that all components must instantiate its own DmssAPI.

Refs #174 (can't close it before I've wrapped the create-dm-app in DMSSProvider)


packages/dm-core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@development-framework/dm-core",
-  "version": "1.0.49",
+  "version": "1.0.50",
```
KristianKjerstad, line 3: might need to be updated to 1.0.51

KristianKjerstad (approved):

PR #186, opened 2023-04-20
fix: avoid eslint failure

ingeridhellen:

Updates the pre-commit hooks and fixes code according to it suggestions. Also removes the eslint config from package.json, since we're only using eslint through the pre-commit and having them there caused some strange errors when building

### DIFF
--- a/packages/dm-core/src/hooks/useRecipe.tsx
+++ b/packages/dm-core/src/hooks/useRecipe.tsx
@@ -97,9 +97,8 @@ export const useRecipe = (
     isLoading: isBlueprintLoading,
     error,
   } = useBlueprint(typeRef)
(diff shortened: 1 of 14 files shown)


packages/dm-core/src/hooks/useRecipe.tsx
```diff
@@ -97,9 +97,8 @@ export const useRecipe = (
     isLoading: isBlueprintLoading,
     error,
   } = useBlueprint(typeRef)
-  const { loading: isPluginContextLoading, getUiPlugin } = useContext(
-    UiPluginContext
-  )
+  const { loading: isPluginContextLoading, getUiPlugin } =
+    useContext(UiPluginContext)
```
soofstad, line 101: Any support for increasing line length limit to 119 as we have in python?

soofstad (approved):

Issue #176, opened 2023-04-06
bug: tree view is not working with the new dmss reference schema for packages


KristianKjerstad:
after we started using this reference schema for pacakges, the tree view is no longer working

```json
       {
            ref: '5b98eff6-3029-4742-9615-fe58aa4a7d14',
            targetName: 'StatusEnum',
            targetType: 'dmss://system/SIMOS/Enum',
            type: 'dmss://system/SIMOS/Link'
        },
```

![Image](https://user-images.githubusercontent.com/69512295/230326971-e4a390a0-5430-4fc8-91a5-7861525ee2b5.png)

KristianKjerstad:
should wait until all bugs from https://github.com/equinor/data-modelling-storage-service/pull/408are fixed

PR #202, opened 2023-05-02
Tree should use new id reference format

eoaksnes:

* Tree should use new id reference format


soofstad (approved):

KristianKjerstad (approved):
also upgrade the version of dm-core?

this might close issue https://github.com/equinor/dm-core-packages/issues/176

PR #206, opened 2023-05-03
ci: test dm-core-plugins with newest dm-core

ingeridhellen:

Use the local dm-core variant when testing packages

When updating dm-core, we had to first merge the changes to dm-core in on PR to get a new npm release, before updating dm-core-plugins to use the newest functionality in another PR due to issues in testing.

Closes #389

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -19,7 +19,7 @@ jobs:
   tests:
     strategy:
       matrix:
-        package: [ dm-core, dm-core-plugins ]
+        package: [dm-core, dm-core-plugins]
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@main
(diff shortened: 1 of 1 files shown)


KristianKjerstad (approved):
👍

.github/workflows/tests.yaml
```diff
@@ -28,12 +28,34 @@ jobs:
...
+      - name: 'Install dependencies'
+        working-directory: packages/${{ matrix.package }}
+        run: yarn install
+
+      - name: 'Copy local variant of dm-core to node_modules'
+        working-directory: packages/${{ matrix.package }}
+        if: steps.variables.outputs.DM_CORE_VERSION != 'null'
+        run: |
+          cd ../dm-core
+          yarn install --no-lockfile
```
soofstad, line 47: yarn install is already done at line 40.

soofstad (approved):

.github/workflows/tests.yaml
ingeridhellen: Yes, but that installs the dependencies in dm-core-plugins. This one installs it in dm-core. We need both

PR #208, opened 2023-05-04
Fix/attribute selector recursiveness

### LINKED ISSUE #204: refactor: use yaml plugin as default for self when no attribute is specified in config
If not the plugin will go into forever...

soofstad:
- Set "yaml" as default plugin for "home" tab in "attribute-selector"
- Check for available docker-compose command in `example/reset-all.sh`
- Rename some test blueprints

- Avoid recursive "attribute-selector"
- Make `reset-all.sh` work with compose as a plugin
- Follow naming conventions
closes #204

### DIFF
--- a/example/reset-all.sh
+++ b/example/reset-all.sh
@@ -1,7 +1,16 @@
 #!/usr/bin/env bash
 set -e
-docker-compose run --rm dmss reset-app
-docker-compose run --rm job-api dm -u http://dmss:5000 reset ../app
+
+# Get the available docker-compose command
+docker compose &> /dev/null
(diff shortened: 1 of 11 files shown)


example/reset-all.sh
```diff
@@ -1,7 +1,16 @@
 #!/usr/bin/env bash
 set -e
-docker-compose run --rm dmss reset-app
-docker-compose run --rm job-api dm -u http://dmss:5000 reset ../app
+
+# Get the available docker-compose command
```
KristianKjerstad, line 4: cool!

KristianKjerstad (approved):
looks good!

PR #213, opened 2023-05-09
fix: wrong references to plugins

eoaksnes:

* Update signal demo app to use correct plugin blueprints

### DIFF
--- a/example/app/data/DemoDataSource/apps/MySignalApp/package.json
+++ b/example/app/data/DemoDataSource/apps/MySignalApp/package.json
@@ -18,12 +18,6 @@
         "protocol": "dmss"
       },
       {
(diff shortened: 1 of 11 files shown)


example/app/data/DemoDataSource/apps/MySignalApp/package.json
```diff
@@ -18,7 +18,7 @@
         "protocol": "dmss"
       },
       {
-        "alias": "PluginModels",
+        "alias": "PLUGINS",
         "address": "DemoDataSource/apps/MySignalApp/models/plugins",
```
soofstad, line 22: Should not "PLUGINS" point to the global standard "system/PLUGINS"?

soofstad (approved):

PR #217, opened 2023-05-24
feature(generic-list): support for pagination

### LINKED ISSUE #109: feat: show list of children using a ui recipe specified in the child

soofstad:
- Adds pagination to "generic-list"-plugin

Screenshot_20230524_132650

closes #109

### DIFF
--- a/packages/dm-core-plugins/src/generic-list/Components.tsx
+++ b/packages/dm-core-plugins/src/generic-list/Components.tsx
@@ -42,3 +42,16 @@ export const MoveItemDownButton = (props: { onClick: () => void }) => (
     <Icon data={arrow_down} title="Move down" />
   </Button>
 )
+
+export const ItemsPerPageButton = (props: {
+  onClick: (n: number) => void
+  amount: number
+  current: number
+}) => (
+  <Button
+    onClick={() => props.onClick(props.amount)}
+    variant={props.current === props.amount ? 'outlined' : 'ghost'}
+  >
+    {props.amount}
+  </Button>
+)
--- a/packages/dm-core-plugins/src/generic-list/GenericListPlugin.tsx
+++ b/packages/dm-core-plugins/src/generic-list/GenericListPlugin.tsx
@@ -8,7 +8,13 @@ import {
   ViewCreator,
   ErrorResponse,
 } from '@development-framework/dm-core'
-import { Button, Icon, Typography } from '@equinor/eds-core-react'
+import {
+  Button,
+  Icon,
+  Typography,
+  Pagination,
+  Label,
+} from '@equinor/eds-core-react'
 import { chevron_down, chevron_right } from '@equinor/eds-icons'
 import { AxiosResponse, AxiosError } from 'axios'
 import React, { useEffect, useState } from 'react'
@@ -16,6 +22,7 @@ import styled from 'styled-components'
 import {
   AppendButton,
   DeleteButton,
+  ItemsPerPageButton,
   MoveItemDownButton,
   MoveItemUpButton,
   SaveButton,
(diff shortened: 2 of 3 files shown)


packages/dm-core-plugins/src/generic-list/GenericListPlugin.tsx
```diff
@@ -38,6 +45,22 @@ const Table = styled.table`
...
+): {
+  [key: string]: TGenericObject
+} {
+  const firstItem = currentPage * itemsPerPage
+  const lastItem = firstItem + itemsPerPage
+  const itemKeys = Object.keys(allItems).slice(firstItem, lastItem)
+  const newItems = Object.values(allItems).slice(firstItem, lastItem)
+  return Object.fromEntries(
+    itemKeys.map((key: string, index: number) => [key, newItems[index]])
+  )
```
ingeridhellen, line 61: You could instead do this, if you like:
```suggestion
  return Object.fromEntries(Object.entries(allItems).slice(firstItem, lastItem)) 
```

packages/dm-core-plugins/src/generic-list/GenericListPlugin.tsx
```diff
@@ -161,98 +190,153 @@ export const GenericListPlugin = (
...
+          />
+        </div>
+        <div
+          style={{
+            display: 'flex',
+            flexDirection: 'row',
+            gridColumnStart: '3',
+            justifySelf: 'end',
+          }}
+        >
```
ingeridhellen, line 332: Could you extract this into a separate component in its own file?

packages/dm-core-plugins/src/generic-list/Components.tsx
```diff
@@ -42,3 +42,16 @@ export const MoveItemDownButton = (props: { onClick: () => void }) => (
     <Icon data={arrow_down} title="Move down" />
   </Button>
 )
+
+export const ItemsPerPageButton = (props: {
+  onClick: (n: number) => void
+  setValue: number
```
ingeridhellen, line 48: Perhaps?:
```suggestion
  setItemsPerPage: (n: number) => void
  value: number
```
Usually in react, a variable whose name starts with "set" is a function.

ingeridhellen (approved):
Looks very nice! Some refactoring suggestions, though

packages/dm-core-plugins/src/generic-list/GenericListPlugin.tsx
soofstad: Much better! Fixed :+1:

packages/dm-core-plugins/src/generic-list/GenericListPlugin.tsx
soofstad: Replaced with @awesthouse suggestion

soofstad: Going to be a bit harder to get all the stuff aligned on one line. Also no that crucial  now that the "ItemsPerPage" buttons got more compact.

packages/dm-core-plugins/src/generic-list/Components.tsx
soofstad: Renamed to "amount" :+1:

packages/dm-core-plugins/src/generic-list/GenericListPlugin.tsx
```diff
@@ -161,98 +190,153 @@ export const GenericListPlugin = (
...
+        }}
+      >
+        <div
+          style={{
+            display: 'flex',
+            flexDirection: 'row',
+            alignItems: 'center',
+          }}
+        >
+          <Label label={'Items per page:'} />
```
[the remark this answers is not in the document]
soofstad: Thats how this should be done :man_facepalming: 
Thanks :smile:

PR #218, opened 2023-05-25
Refactor/yarn workspaces

awesthouse:
- Implements [yarn workspaces](https://classic.yarnpkg.com/lang/en/docs/workspaces/) in project
   - Creates monorepo of `example`, `dm-core` and `dm-core-plugins`
   - Removes postbuild from `dm-core` and `dm-core-plugins` build
   - Updates example to use yarn workspaces linking of repos instead of yarn link or symlink
- Fixes dependency issues between projects. Removed peerDependencies from dependencies and devDependencies when they were defined both places both places
- Updates documentation
- Adds a performance friendly way to have hot reload and listen to changes when running example and editing in `dm-core` and `dm-core-plugins` by using aliases that switches out package with use of index files in package folder.

- Simplifies initial setup of project (only one yarn install)
- Allows user to run all commands from main project folder (we can add more commands as we see fit)
- Removes the need to delete node_modules folders in dm-core and dm-core-plugins
- Fixes peer-dependency issues (also related to the point above)
- Updates documentation with missing steps

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -19,7 +19,7 @@ jobs:
   tests:
     strategy:
       matrix:
-        package: [ dm-core, dm-core-plugins ]
+        package: [dm-core, dm-core-plugins]
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@main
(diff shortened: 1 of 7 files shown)


awesthouse:
Just a suggestion 🙂 
Obviously have to update to fit build steps and stuff, but wanted to hear what you think and give option to test locally

KristianKjerstad:
I think this is a good suggestion, but  it's difficult to test if all parts work when DMSS and is not working with this repo and / or dm-cli.

Perhaps we should wait with this change until after we get a new stable version of dmss (https://github.com/equinor/data-modelling-storage-service/issues/414 is closed)

.github/workflows/tests.yaml
```diff
@@ -28,34 +28,11 @@ jobs:
...
-      #        run: |
-      #          cd ../dm-core
-      #          yarn install --no-lockfile
-      #          yarn build
-      #          cd ../dm-core-plugins
-      #          rm -r node_modules/@development-framework/dm-core/*
-      #          cp -r ../dm-core/dist node_modules/@development-framework/dm-core/dist
-      #          cp ../dm-core/package.json node_modules/@development-framework/dm-core/package.json
-
-      - name: 'yarn tsc'
```
eoaksnes, line 54: I guess this should still be here, to test that the typescript works?

eoaksnes (approved):
Very nice! :)

.github/workflows/tests.yaml
awesthouse: Build in the individual packages is `yarn tsc`, wouldn't that be the same? 🙂

PR #224, opened 2023-06-02
fix: loading bug in tree

### LINKED ISSUE #222: fix: add proper loading icon to tree view in dm-core-plugins

KristianKjerstad:
Before, the call to "expand tree" was buggy and did not do things in correct order. This issue is now fixed

Also, update some API calls that were wrong

closes #222

### DIFF
--- a/packages/dm-core/src/domain/Tree.tsx
+++ b/packages/dm-core/src/domain/Tree.tsx
@@ -211,7 +211,7 @@ export class TreeNode {
 
   expand = () => {
     return new Promise((resolve, reject) => {
-      ;(async () => {
+      return (async () => {
         if (!this.isDataSource) {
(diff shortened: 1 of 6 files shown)


packages/dm-core/src/domain/Tree.tsx
```diff
@@ -209,51 +209,67 @@ export class TreeNode {
...
-          this.tree.index[this.dataSource].type = 'error'
-          this.tree.index[this.dataSource].message = error.message
-        })
-        .finally(() => {
-          this.tree.updateCallback(this.tree)
-        })
-    }
+  expand = () => {
+    return new Promise((resolve, reject) => {
+      ;(async () => {
```
ingeridhellen, line 214: What syntax is this? Starting the line with `;`?

ingeridhellen (approved):
Nice!

Issue #220, opened 2023-06-01
fix: make signal app and plugins in dm-core-package bug free


KristianKjerstad:
# Tasks

# Completed
- [x] Tree component gives error `Required parameter typeRef was null or undefined when calling blueprintGet.` when opening 
clicking on a "sub attribute" of an entity, for example the listOfObjects inside formExample @KristianKjerstad  https://github.com/equinor/dm-core-packages/pull/305 

![Image](https://github.com/equinor/dm-core-packages/assets/69512295/a041eeba-234e-4847-a920-650de92142c1)

- [x] open newly created case gives error @KristianKjerstad  #286
How to reproduce:
1) click "+ Add row"
2) click on the + button to open the newly created row.
![Image](https://github.com/equinor/dm-core-packages/assets/69512295/61361086-3f22-4718-9dee-c4289ac434f0)

- [x] deleting a newly created case gives 404 error message

- [x] attribute selector tabs example (under attribute selector plugin) gives errors because it has not created all non-optional attributes @KristianKjerstad  https://github.com/equinor/dm-core-packages/pull/278

 

- [x] open inputEnity does not work @KristianKjerstad  https://github.com/equinor/dm-core-packages/pull/249

![Image](https://user-images.githubusercontent.com/1190419/239824936-8738012a-4c26-4340-8b16-d7f4dbe48377.png)

- [x] Rename SideBar to ???? @KristianKjerstad 

![Image](https://user-images.githubusercontent.com/1190419/239822468-6319e3bd-e5c9-4221-a023-86ac8178d45e.png)

- [x] Not able to add cases in Signal app demo  @KristianKjerstad  https://github.com/equinor/data-modelling-storage-service/compare/fix/add-document-bug

![Image](https://github.com/equinor/dm-core-packages/assets/1190419/79ffab20-1cb5-4b1d-86d2-b2fb6d011d40)

- [x] update dm-job with the latest dm-cli package and fix the $ when fetching id @eoaksnes
- [x] submit (save) a case fails 

![Image](https://user-images.githubusercontent.com/1190419/239821628-4bdca185-519a-427e-8797-cd4bedfdc144.png)

- [x] - [ ] Add a new cace through UI fails

![Image](https://github.com/equinor/dm-core-packages/assets/69512295/790487e7-32a5-492e-8a25-62c153ae270b)

[...]

PR #227, opened 2023-06-05
Feat/signal components

awesthouse:
- Added signal plugins from Babak. 
- Added missing initialUiPlugins in CaseUI

SignalPlot has been somewhat rewritten (simple cleanup). 
~~Will rewrite components some more and remove material-ui and more unnecessary code in later PR - it's just wanted that we add this first for testing purposes.~~

Update:
- Rewritten SignalTable to use eds
- Removed material-ui packages
- Removed unused style variables and components

Add Signal app plugins

#220


example/package.json
```diff
@@ -3,12 +3,16 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
+    "@craco/craco": "^7.0.0",
```
eoaksnes, line 6: Should this be in devDependencies?

eoaksnes (approved):
Nice! Just a single comment about craco npm package, if it should be moved to devDeps.

PR #243, opened 2023-06-09
Refactor/form plugin spacing

awesthouse:
- Introduced Stack (spacer/flexbox) component to dm-core
- Removed margins from specific elements
- Removed duplicate wrapper components
- Used Stack component to create spacing in form-plugin

- Remove individual margins
- Less duplicate styling/wrapper components
- More uniform spacing of elements


eoaksnes (approved):

eoaksnes:
Nice, this will be helpful! :)

packages/dm-core-plugins/src/form/fields/ArrayField.tsx
```diff
@@ -45,7 +26,7 @@ export default function Fields(props: any) {
   })
 
   const handleAddObject = () => {
-    const name: string = `${namePath}-${fields.length}`
+    // const name: string = `${namePath}-${fields.length}`
```
KristianKjerstad, line 29: can comment be removed?

packages/dm-core/src/components/common/Stack/Stack.tsx
```diff
@@ -0,0 +1,28 @@
...
+      flexDirection={direction}
+      flexGrow={grow}
+      flexShrink={shrink}
+      flexWrap={wrap}
+      {...restProps}
+    />
+  )
+}
+
+Stack.defaultProps = {
```
KristianKjerstad, line 18: cool! Was not aware of this syntax to define default prop values

KristianKjerstad (approved):
nice!

Also, I guess you can update versions of dm-core/package.json and dm-core-plugins/package.json to release the changes to npm

packages/dm-core-plugins/src/form/fields/ArrayField.tsx
awesthouse: Yeah! I just didn't know what it was for and it showed as a warning so I commented it out, but we might as well delete it since it was unused 🙂

PR #233, opened 2023-06-07
Fix/form plugin bugs

KristianKjerstad:
## Issues related to this change
https://github.com/equinor/dm-core-packages/issues/220

PR #246, opened 2023-06-09
Feat/make entity picker button able to pick reference to entity

KristianKjerstad:

Update some dm-core components to handle the new reference format.

required before https://github.com/equinor/dm-core-packages/pull/233 can be merged

goal: be able to select a reference with the entity picker button.

#220

### DIFF
--- a/packages/dm-core/src/components/NewEntityButton.tsx
+++ b/packages/dm-core/src/components/NewEntityButton.tsx
@@ -11,7 +11,7 @@ import {
 import { addToPath } from './UploadFileButton'
 import { AxiosError } from 'axios'
 import styled from 'styled-components'
-import { TReference } from '../types'
+import { TGenericObject, TReference } from '../types'
 import { INPUT_FIELD_WIDTH } from '../utils/variables'
 import { useDMSS } from '../context/DMSSContext'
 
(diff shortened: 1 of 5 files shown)


awesthouse (approved):
Still unsure about some context 😅 but code looks good!

packages/dm-core/src/components/NewEntityButton.tsx
```diff
@@ -143,21 +146,25 @@ export function NewEntityButton(props: {
...
               onClick={() => {
                 setLoading(true)
 
-                if (copyTarget) {
+                if (documentToCopy) {
                   // @ts-ignore
-                  delete copyTarget._id
-                  copyTarget.name = newName
+                  delete documentToCopy._id
+                  documentToCopy.name = newName
```
ingeridhellen, line 162: Shouldn't state variables be immutable? I don't think it will matter in this case, but as a principle it would be better to make a copy
```suggestion
                  const newDocument = { ...documentToCopy, name: newName }
                  delete newDocument._id
```
Also, I don't think the @ts-ignore is doing anything anymore

packages/dm-core/src/components/NewEntityButton.tsx
```diff
@@ -11,7 +11,7 @@ import {
 import { addToPath } from './UploadFileButton'
 import { AxiosError } from 'axios'
 import styled from 'styled-components'
-import { TReference } from '../types'
+import { TGenericObject, TReference } from '../types'
```
ingeridhellen, line 14: One place where TReference is used, at least

ingeridhellen (commented):
Looks good

packages/dm-core/src/components/NewEntityButton.tsx
KristianKjerstad: agree! Will fix

PR #255, opened 2023-06-19
refactor: restructure folders for forms

roar-larsen:
Restructuring the folders for plugin-forms to make it easier to keep it readable and separate the different cases. There is now a folder for each case, where also the blueprint is stored.

SingleField is deleted as it is no longer needed or used.

The relevant E2E-tests have been updated to reflect the changes.


KristianKjerstad (approved):
looks good!

PR #257, opened 2023-06-20
refactor: signalapp recipes

ingeridhellen:
## What does this pull request change and why is this pull request needed?

- The signal app recipes contained several uirecipes that were never used. This creates unnecessary clutter. I have removed most of those, if not all.
- The component.json file was renamed to SinCompUI.json, to better align with the naming standard of the other recipes
- Adapted grid spacing to avoid a lot of whitespace at the bottom of the screen when opening a case


eoaksnes (approved):
Nice cleanup :)

PR #264, opened 2023-06-22
Feat/rewrite table and list plugins

### LINKED ISSUE #258: feat: improve table plugins
- [x] Map from objects to array instead of mapping from objects to objects with array values
- [x] Add option to open in new tab
- [x] Add config to allow user to select between open in new tab or inline
- [x] Add pagination (dm-core)
- [x] Fix markup error when in non-edit mode
- [x] Possibility to define open or expand, both, or neither, actions in configuration, plus their relevant views.
- [x] In the column headers, use "label" (or title?) when specified in blueprint. If not specified then use "name".
- [x] Add pagination button titles @roar-larsen #366
- [x] When you submit the form, two instances are stored. Same happens when you edit an item. https://github.com/equinor/data-modelling-storage-service/pull/512 @KristianKjerstad 
- [x] Fix bug with pagination: If you add enough items so that pagination becomes active, you can now paginate beyond the number of items in the table: 
Image
- [ ] when save, only update the rows that have changes

awesthouse:
- feat: adds ability to pass rootId as parameter in addView function and open subviews to AttributeSelector
- chore: updates equinor font source in signal app to new source
- feat: adds pagination component to dm-core
- refactor: rewrite of table and list plugins

**List plugin**
- Removed table markup, now just printed as a list of items
- Uses functionality config object instead of individual value fields
- Can open in new tab
- Rewritten from mapping items from object to object with array value, to mapping to array
- Switched out pagination to dm-core pagination

**Table plugin**
- Cleaned up wrappers
- Uses functionality config object instead of individual value fields
- Can open in new tab and expand view
- Rewritten from mapping items from object to object with array value, to mapping to array
- Added pagination to table component
- Allow user to configure which columns are editable when edit is activated

- Clean up of components
- Expanding functionality (pagination, edit, expand/open tab)
- Make Table and list plugin more different so we don't have two plugins that are almost identical

Closes #258


packages/dm-core-plugins/src/generic-list/GenericListPlugin.tsx
```diff
@@ -161,170 +147,144 @@ export const GenericListPlugin = (
...
+                      ? 'Minimize'
+                      : 'Expand'
+                    : 'Open in new tab'
+                }
+              >
+                <Button
+                  variant="ghost_icon"
+                  color="secondary"
+                  onClick={
+                    internalConfig.openInline
```
KristianKjerstad, line 192: I think there is a bug related to this button click

when clicking the + icon several times for the same case, several tabs will open. 
I guess the wanted functionality is to jump to the already opened tab.

image

packages/dm-core-plugins/src/generic-list/GenericTablePlugin.tsx
```diff
@@ -1,95 +1,153 @@
...
-  const [isSaveLoading, setIsSaveLoading] = useState<boolean>(false)
-  const [dirtyState, setDirtyState] = useState<boolean>(false)
   const dmssAPI = useDMSS()
+  const paginatedRows = items.slice(
+    paginationPage * rowsPerPage,
+    paginationPage * rowsPerPage + rowsPerPage
+  )
+
+  console.log(config)
+  console.log(functionality)
```
KristianKjerstad, line 87: remove console.log()?

packages/dm-core-plugins/src/generic-list/GenericTablePlugin.tsx
```diff
@@ -98,121 +156,181 @@ export const GenericTablePlugin = (
...
+              <Table.Cell width="80"></Table.Cell>
             )}
-            {internalConfig.columns.map((attribute: string) => (
+            {config.columns.map((attribute: string) => (
               <Table.Cell key={attribute}>{attribute}</Table.Cell>
             ))}
-            {internalConfig.editMode && internalConfig.showDelete && (
-              <Table.Cell>Delete</Table.Cell>
+            {functionality?.delete && (
+              <Table.Cell width="80">Delete</Table.Cell>
```
KristianKjerstad, line 193: This is more of a personal preference, but it might make sense to only show attribute name in headers (and leave this cell empty), since it is very clear from the symbol to understand that the button will do a delete.
You can consider leabing this cell blank :) 

image

packages/dm-core-plugins/src/attribute-selector/AttributeSelectorPlugin.tsx
```diff
@@ -31,14 +31,14 @@ export const AttributeSelectorPlugin = (
   const [views, setViews] = useState<TItemData[]>([])
   const [formData, setFormData] = useState<TGenericObject>({})
 
-  function addView(viewId: string, view: TViewConfig) {
+  function addView(viewId: string, view: TViewConfig, rootId?: string) {
```
KristianKjerstad, line 34: could not find anywhere in the changed code that addView() is used. Could you double check if the rootId is needed? :)

packages/dm-core/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,77 @@
...
+  const { count = 0, page, setPage, rowsPerPage, setRowsPerPage } = props
+
+  const getVisibleToLabel = () => {
+    if (count === -1) {
+      return (page + 1) * rowsPerPage
+    }
+    return rowsPerPage === -1
+      ? count
+      : Math.min(count, (page + 1) * rowsPerPage)
+  }
```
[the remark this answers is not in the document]
awesthouse: Fair point! I took a look at how mui solved this 'cause I had a hard time conceptualizing what we actually needed to print and this was their solution, but I'll take another look at this 👍🏻

example/app/data/DemoDataSource/recipes/apps/signal_app/recipe_links/signals_simple/CaseUI.json
```diff
@@ -96,13 +96,19 @@
     {
       "name": "List",
       "type": "CORE:UiRecipe",
-      "plugin": "@development-framework/dm-core-plugins/generic-list",
+      "plugin": "@development-framework/dm-core-plugins/generic-table",
```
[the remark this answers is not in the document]
awesthouse: Yeah, I'll do a separate refactoring pr cause we need to rename the plugin, components and config types

packages/dm-core-plugins/src/generic-list/GenericListPlugin.tsx
awesthouse: I mentioned this very briefly yesterday I think, but I also feel like this is a separate PR 'cause it's more related to the attribute-selector and how we add views 👍🏻 I have plans to fix it though

packages/dm-core-plugins/src/generic-list/GenericTablePlugin.tsx
awesthouse: Yep! I left this as it was, but I also think it makes sense to leave it blank 👍🏻

packages/dm-core-plugins/src/attribute-selector/AttributeSelectorPlugin.tsx
awesthouse: Of course! addView is passed as the onOpen function in <Content />, so in both Table and List component, the onOpen function we call in openItemAsTab function is actually addView, and there we pass the rootId 🙂

packages/dm-core/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,77 @@
+import React from 'react'
+import { Stack } from '../common'
+import { Button, Icon, NativeSelect, Typography } from '@equinor/eds-core-react'
+import { chevron_left, chevron_right } from '@equinor/eds-icons'
+
+type PaginationProps = {
+  count: number
+  page: number
+  setPage: React.Dispatch<React.SetStateAction<number>>
+  rowsPerPage: number
```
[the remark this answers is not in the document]
awesthouse: Actually let me look at this in a separate pr and let user define what options they want in the select 🙂 We'll just have 5 | 10 | 25 | 50 | 100 | 500 as a default maybe

KristianKjerstad (approved):
nice! 👍

PR #267, opened 2023-06-23
feat: select next, previous then self on view/tab close

awesthouse:
When closing a tab (view), instead just selecting "self": select next tab if it exists, select previous tab if it doesn't, select self if the previous two aren't an option. This seems to be the default in browsers.

Feels unnatural that it always goes to self.


ingeridhellen (approved):
I agree!

PR #268, opened 2023-06-23
fix: renaming nested form example

roar-larsen:
Renamed two files and updated their references in other files.

A recipe named "nested" is needed for another case. Therefore had to rename nested.json


example/app/data/DemoDataSource/plugins/form/example/blueprints/NestedField.json
ingeridhellen, line 1: This file should also be renamed

KristianKjerstad (approved):

ingeridhellen (approved):

PR #269, opened 2023-06-23
fix: publish packages using root folder

awesthouse:
Try to publish packages using root folder and yarn workspaces

Run currently fails because dm-core-plugins can't locate dm-core after removing linking of packages

### DIFF
--- a/.github/workflows/on-main-push.yaml
+++ b/.github/workflows/on-main-push.yaml
@@ -21,11 +21,11 @@ jobs:
       job_version: ${{ needs.integration-tests.outputs.job_version }}
 
   publish-packages:
-    needs: [ tests ]
+    needs: [tests]
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        package: [ dm-core, dm-core-plugins ]
+        package: [dm-core, dm-core-plugins]
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/on-main-push.yaml
```diff
@@ -50,9 +50,9 @@ jobs:
       - name: Create package distribution and publish
         if: steps.check-version.outputs.NEW_VERSION == 'true'
         run: |
-          cd packages/${{ matrix.package }}
           yarn install
-          yarn build
-          yarn publish --access public
+          yarn build:packages
+          yarn workspace dm-core publish --access public
+          yarn workspace dm-core-plugins publish --access public
```
ingeridhellen, line 56: I thought it were
```suggestion
          yarn workspace @development-framework/dm-core publish --access public
          yarn workspace @development-framework/dm-core-plugins publish --access public
```
? Don't really know this syntax well though

ingeridhellen (approved):
Great!

.github/workflows/on-main-push.yaml
awesthouse: I think you're right yeah! Thanks 👍🏻 Hard to test without committing and seeing how it runs 😅

PR #271, opened 2023-06-23
🪟 Docs/Readme for Windows Setup🪟

pbullhove:

This pull request adds a step-by-step setup guide for Windows users to start the DMT Example application. 
It is a list of commands. 

It is tested that these steps works on a Windows computer. 

🪟 vs 🍎
The startup steps are not the same as for mac, and therefore the current step-by-steps did not work. Some people will need to access the application from Windows computers in the future, and therefore this guide is required. 

Not connected to an issue.

### DIFF
--- a/example/README.md
+++ b/example/README.md
@@ -50,4 +50,7 @@ docker-compose down --volumes && docker-compose pull && docker-compose up -d && 
 4. Make sure you have pip and venv installed in the WSL (Ubuntu) system. 
    1. `sudo apt-get upgrade && sudo apt-get upgrade`
    2. `sudo apt install python3-pip && sudo apt install python3-venv`
-
+5. In the WSL terminal navigate to `dm-core-packages/example` and run `python3 -m venv .venv`. This will take a few minutes to run. Now grab a coffee. ☕
+6. Run `source .venv/bin/activate`
+7. Run `pip install dm-cli && dm reset app && dm import-plugin-blueprints node_modules/@development-framework/dm-core-plugins && dm create-lookup example DemoDataSource/recipes`
+8. Using the Powershell terminal, navigate to root in `dm-core-packages` and start the app with the command `yarn start:example`. The app should now be served on `localhost:5000`. 
--- a/example/README.md
+++ b/example/README.md
@@ -40,17 +40,20 @@
 
 ## Running (Windows)
 
-1. Open Powershell and navigate to dm-core-packages/example
+1. Open Powershell and navigate to dm-core-packages/example.
 2. Run
 ```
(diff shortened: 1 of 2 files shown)


example/README.md
```diff
@@ -37,3 +37,38 @@
    - `./reset-all.sh`
 5. Start the test app
    - `yarn start:example`
+
+## Running (Windows)
+
+1. Open Powershell and navigate to dm-core-packages/example.
+2. Make sure docker engine is running, then run these commands
+   - `yarn install`
```
awesthouse, line 45: yarn install should be run from dm-core-packages folder

example/README.md
```diff
@@ -37,3 +37,38 @@
...
+   - `docker-compose down --volumes`
+   - `docker-compose pull`
+   - `docker-compose up -d`
+   - `docker-compose run --rm dmss reset-app` (This fails sometimes, try again if it fails. )
+   - `docker compose run --rm job-api dm -u http://dmss:5000 reset ../app`
```
awesthouse, line 50: Do you not need to install dm-cli _before_ running this? Would it make sense to add steps to install venv and dm-cli before saying running docker-compose pull and up? 🙂

ingeridhellen: I don't think it matters what you do first of installing dm-cli and running the dc commands. The dc commands run in their own environment, and don't require anything of you computer setup once you have a dc file

awesthouse: I just have a memory of not being able to run these commands because I hadn't installed dm-cli and/or installed and acivated venv - but I might be remembering wrong 🙂

example/README.md
pbullhove: Yarn install is now from the correct folder, thanks for spotting that. 

Does not need yarn build:dm-core and dm-core-plugins. 
Was not able to get the commands to run without errors, but it works nonetheless without building. And Andrea said it should work without building the packages locally.

example/README.md
pbullhove: Do not need to install dm-cli before this.

package.json
```diff
@@ -12,5 +12,8 @@
     "clear-external-pckgs": "rm -rf node_modules && rm -rf packages/dm-core/node_modules && rm -rf packages/dm-core-plugins/node_modules && rm -rf example/node_modules && rm yarn.lock",
     "build:packages": "yarn workspace @development-framework/dm-core build && yarn workspace @development-framework/dm-core-plugins build",
     "test:packages": "yarn workspace @development-framework/dm-core test && yarn workspace @development-framework/dm-core-plugins test"
+  },
+  "dependencies": {
+    "yarn": "^1.22.19"
```
awesthouse, line 17: This shouldn't be here 🙂

awesthouse (changes requested):
README looks good, but yarn shouldn't be a dependency so it has to be removed before merging

awesthouse (approved):
💯

PR #273, opened 2023-06-23
refactor: don't store eds icons in static file, use eds-icons package

awesthouse:
Uses EDS icons directly from icons-package and removes dependency to static file 

Don't need to handle and update static file of eds-icons


ingeridhellen (approved):

ingeridhellen:
Nice!

PR #275, opened 2023-06-23
feat: enable custom form widgets

ingeridhellen:
## What does this pull request change, and why?

I've removed some unnecessary folder levels and index files from form/widgets
I've moved two widgets that were defined in <Form/> into separate files
I've created a widget context and removed widget logic from the registrycontext. The widget context can be wrapped around an application to allow for custom widgets

Refs #266

### DIFF
--- a/packages/dm-core-plugins/src/form/widgets/TypeWidget.tsx
+++ b/packages/dm-core-plugins/src/form/widgets/TypeWidget.tsx
@@ -4,7 +4,7 @@ import { TWidget } from '../types'
 import TextWidget from './TextWidget'
 
 const TypeWidget = (props: TWidget) => {
-  const { id, namePath, label, value } = props
+  const { id, label, value } = props
   const { blueprint, isLoading, error } = useBlueprint(value)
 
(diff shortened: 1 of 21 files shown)


KristianKjerstad (approved):
Good changes!

I think @eoaksnes also would like to review since he knows the form plugin the best

packages/dm-core-plugins/src/form/widgets/TypeWidget.tsx
```diff
@@ -0,0 +1,36 @@
...
+        id={id}
+        label={label}
+        readOnly={true}
+        value={value}
+        // eslint-disable-next-line @typescript-eslint/no-empty-function
+        onChange={() => {}}
+        onClick={() => {
+          // @ts-ignore
+          window
+            .open(`dmt/view/${datasourceId}/${blueprint.uid}`, '_blank')
```
eoaksnes, line 28: Since this does not work, should we just remove the .open?

eoaksnes (approved):
Looks good, nice cleanup!

PR #277, opened 2023-06-26
fix: expanded simple-form pluging and updated e2e

roar-larsen:
Expanded the simple form plugin to include all types of primitives, with some variations.

The e2e-test have been updated. Note that some lines have been commented out because there is a bug in the form package which will cause the test to fail and we don´t want this test to stop other work for now. See [https://github.com/orgs/equinor/projects/189?pane=issue&itemId=31708452](https://github.com/orgs/equinor/projects/189?pane=issue&itemId=31708452 )

### DIFF
--- a/e2e/tests/plugin-form-Simple.spec.ts
+++ b/e2e/tests/plugin-form-Simple.spec.ts
@@ -9,7 +9,7 @@ test('Simple form', async ({ page }) => {
   await page.getByText('DemoDataSource/$Simple').click()
 
   //Remove prefilled optional string
-  await page.getByLabel('Optional string (optional)').fill('')
+  await page.getByLabel('An optional string (optional)').fill('')
 
   //Fill out required string
   await page.getByTestId('form-submit').click()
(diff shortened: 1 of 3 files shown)


e2e/tests/plugin-form-Simple.spec.ts
```diff
@@ -8,30 +8,48 @@ test('Simple form', async ({ page }) => {
   await page.getByText('simple', { exact: true }).click()
   await page.getByText('DemoDataSource/$Simple').click()
 
-  //Fill out optional string
-  await page.getByLabel('Optional string (optional)').fill('Foo')
+  //Remove prefilled optional string
+  await page.getByLabel('Optional string (optional)').fill('')
```
ingeridhellen, line 12:
```suggestion
  await page.getByLabel('An optional string (optional)').fill('')
```

example/app/data/DemoDataSource/plugins/form/simple/simple.json
```diff
@@ -2,5 +2,5 @@
   "_id": "Simple",
   "name": "Simple",
   "type": "./blueprints/Simple",
-  "Firstname": ""
+  "stringOptional": "a prefilled optional string"
```
ingeridhellen, line 5: This does not seem to be a correct implementation of Simple. It lacks the non-optional attributes `stringRequired` and `requiredCheckbox`, and it contains `name`. which is not defined in the blueprint.

ingeridhellen (approved):

PR #278, opened 2023-06-26
fix: better error handing in attriubte selector

KristianKjerstad:

add error handling to attribute selector plugin. 

Do not crash if a single attribute is not found

requires https://github.com/equinor/data-modelling-storage-service/pull/456

#220

### DIFF
--- a/packages/dm-core-plugins/src/attribute-selector/Content.tsx
+++ b/packages/dm-core-plugins/src/attribute-selector/Content.tsx
@@ -22,7 +22,6 @@ export const Content = (props: {
   formData: TGenericObject
 }): JSX.Element => {
   const { selectedView, items, setFormData, formData, onOpen, type } = props
-
   return (
     <div style={{ width: '100%' }}>
       {items.map((config: TItemData) => (
@@ -30,7 +29,6 @@ export const Content = (props: {
           key={config.viewId}
           hidden={config.viewId !== selectedView}
         >
-          {/*@ts-ignore*/}
           <ViewCreator
             idReference={config.rootEntityId}
             viewConfig={config.view}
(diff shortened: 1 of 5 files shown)


packages/dm-core-plugins/src/attribute-selector/Content.tsx
```diff
@@ -22,19 +22,19 @@ export const Content = (props: {
...
         <HidableWrapper
           key={config.viewId}
           hidden={config.viewId !== selectedView}
         >
-          {/*@ts-ignore*/}
           <ViewCreator
             idReference={config.rootEntityId}
             viewConfig={config.view}
-            onOpen={onOpen}
+            onOpen={(id: string, v: TViewConfig) => {
```
ingeridhellen, line 35: Why?

packages/dm-core/src/components/ViewCreator/ViewCreator.tsx
```diff
@@ -50,19 +51,25 @@ export const ViewCreator = (props: TViewCreator): JSX.Element => {
...
+      .then((response: AxiosResponse) => {
         setAttribute(response.data)
       })
       .catch((error) => setError(error))
       .finally(() => setIsLoading(false))
   }, [])
 
   if (isLoading) return <Loading />
-  if (error) throw error
+  if (error)
```
ingeridhellen, line 62: I think the proper way to do this would be to make an ErrorBoundary plugin that would catch all errors and display them in a nice way. That way we can be sure to avoid the page crashing in an unexpected way

ingeridhellen (approved):

packages/dm-core/src/components/ViewCreator/ViewCreator.tsx
KristianKjerstad: adding a note about this to https://github.com/orgs/equinor/projects/189/views/3?filterQuery=is%3Aissue+-cardtype%3AEpic+error+bo&pane=issue&itemId=31823204 :)

PR #286, opened 2023-06-27
fix: bug when opening unsaved table row

KristianKjerstad:
fix bug: before, when opening an unsaved row we got this error
image

the bug is now fixed - user have to press save before opening a case

#220


KristianKjerstad:
> Mostly good. However, if you edit an existing item, then open it, it will open without the new changes. Should we disable the "open" button in those cases too?

yeah, I think that makes sense!

PR #288, opened 2023-06-27
refactor: update ts types for blueprint and attr

ingeridhellen:

The typing seems a bit outdated. Now that it's up to date, we can use it more

### DIFF
--- a/packages/dm-core/src/domain/types.ts
+++ b/packages/dm-core/src/domain/types.ts
@@ -1,21 +1,22 @@
 export type TAttribute = {
-  type?: string
   name: string
-  dimensions: string
+  type: string
   attributeType: string
+  dimensions?: string
   optional?: boolean
   contained?: boolean
   description?: string
   label?: string
   enumType?: string
-  default?: string
+  default?: any
 }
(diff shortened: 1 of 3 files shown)


packages/dm-core/src/domain/types.ts
```diff
@@ -1,21 +1,22 @@
...
   name: string
   type: string
-  description: string
-  attributes: TAttribute[]
-  uiRecipes?: any[]
-  storageRecipes?: any[]
+  description?: string
+  extends?: string[]
+  attributes?: TAttribute[]
+  _meta_?: any[]
```
KristianKjerstad, line 20: is _meta_ a TGenericObject?

KristianKjerstad (approved):
👍

packages/dm-core/src/domain/types.ts
ingeridhellen: No. _meta_ has its own blueprint, and on of its attributes has its own blueprint... I didn't see the point in making types for this since I doubt anyone will use the meta attribute in frontend

PR #289, opened 2023-06-27
fix: bugs in generic list plugin

KristianKjerstad:

fix some issues with the generic list plugin and the generic list example:

- the recipe for orderitem was wrong. Before, we tried to use the generic list plugin on a document that was NOT a list (the generic list plugin was used on an entity of type OrderItem)
-  update generic list plugin to have similar behaviour as the generic table plugin: do not call dmss when adding a row (instead, only call dmss when pressing save) and also add the `isSaved` attribute to ItemRow type

#220

### DIFF
--- a/example/app/data/DemoDataSource/recipes/plugins/generic-list/orderItem.json
+++ b/example/app/data/DemoDataSource/recipes/plugins/generic-list/orderItem.json
@@ -21,113 +21,6 @@
             },
             "scope": "self"
           }
-        },
-        {
-          "type": "PLUGINS:dm-core-plugins/attribute-selector/AttributeSelectorItem",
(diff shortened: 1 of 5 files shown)


example/app/data/DemoDataSource/recipes/plugins/generic-list/orderItem.json
```diff
@@ -21,103 +21,6 @@
             },
             "scope": "self"
           }
-        },
```
KristianKjerstad, line 24: The old config was wrong, since the orderItem is NOT a list it does not make sense to define list plugins in the initial UI recipe

ingeridhellen (approved):
Nice!

PR #293, opened 2023-06-28
Refactor/table refactor

awesthouse:
- rename GenericTablePlugin to TablePlugin
- move TablePlugin to own folder (from generic-list folder to table folder)
- Get blueprints in reset-all script from correct node_modules folder 
- Fix of runtime error that appeared suddenly 
- Update recipes that use table plugin with new types and use of functionality object

- Need of name change
- Need to update recipes that were broken when Table component was updated

#258

### DIFF
--- a/example/reset-all.sh
+++ b/example/reset-all.sh
@@ -12,6 +12,6 @@ set -e
 eval $compose run --rm dmss reset-app
 eval $compose run --rm job-api dm -u http://dmss:5000 reset ../app
 dm reset app
-dm import-plugin-blueprints node_modules/@development-framework/dm-core-plugins
+dm import-plugin-blueprints ../node_modules/@development-framework/dm-core-plugins
 echo "Creating recipe lookup..."
 dm create-lookup example DemoDataSource/recipes
--- a/packages/dm-core-plugins/blueprints/table/TableFunctionalityConfig.json
+++ b/packages/dm-core-plugins/blueprints/table/TableFunctionalityConfig.json
@@ -1,0 +1,48 @@
+{
+  "type": "CORE:Blueprint",
+  "name": "TableFunctionalityConfig",
+  "attributes": [
+    {
+      "attributeType": "boolean",
+      "type": "CORE:BlueprintAttribute",
+      "name": "openAsTab",
+      "optional": true,
+      "default": true
+    },
+    {
+      "attributeType": "boolean",
+      "type": "CORE:BlueprintAttribute",
+      "name": "openAsExpandable",
+      "optional": true,
+      "default": false
+    },
+    {
+      "attributeType": "boolean",
+      "type": "CORE:BlueprintAttribute",
+      "name": "delete",
+      "optional": true,
+      "default": false
+    },
+    {
+      "attributeType": "boolean",
+      "type": "CORE:BlueprintAttribute",
+      "name": "add",
+      "optional": true,
+      "default": false
+    },
+    {
+      "attributeType": "boolean",
+      "type": "CORE:BlueprintAttribute",
+      "name": "edit",
+      "optional": true,
+      "default": false
+    },
+    {
+      "attributeType": "boolean",
+      "type": "CORE:BlueprintAttribute",
+      "name": "sort",
+      "optional": true,
+      "default": false
+    }
+  ]
+}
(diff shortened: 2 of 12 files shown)


example/reset-all.sh
```diff
@@ -12,6 +12,6 @@ set -e
 eval $compose run --rm dmss reset-app
 eval $compose run --rm job-api dm -u http://dmss:5000 reset ../app
 dm reset app
-dm import-plugin-blueprints node_modules/@development-framework/dm-core-plugins
+dm import-plugin-blueprints ../node_modules/@development-framework/dm-core-plugins
```
KristianKjerstad, line 15: Don't think this change is necessary since the old script worked if following the README (which specifies that user should run the reset-all script from the example folder.

packages/dm-core-plugins/blueprints/table/TablePluginConfig.json
```diff
@@ -1,12 +1,7 @@
 {
   "type": "CORE:Blueprint",
-  "name": "GenericTablePluginConfig",
+  "name": "TablePluginConfig",
   "attributes": [
-    {
-      "attributeType": "string",
-      "type": "CORE:BlueprintAttribute",
-      "name": "type"
-    },
```
KristianKjerstad, line 9: for all blueprints, type is required, so THis can1t be removed I believe

packages/dm-core-plugins/blueprints/table/TableFunctionalityConfig.json
```diff
@@ -0,0 +1,48 @@
+{
+  "type": "CORE:Blueprint",
+  "name": "TableFunctionalityConfig",
+  "attributes": [
+    {
```
KristianKjerstad, line 5: an attribute with name "type" is required for blueprints :) 

```suggestion
    {
          "attributeType": "string",
          "type": "CORE:BlueprintAttribute",
          "name": "type",
    }
```

packages/dm-core-plugins/src/table/types.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export type ITableItemRow = {
+  key: string
+  data: any
+  index: number
+  expanded: boolean
+  isSaved: boolean
+}
+
+export type TablePluginProps = {
```
KristianKjerstad, line 45:
```suggestion
export type TTablePluginProps = {
```

packages/dm-core-plugins/src/table/types.ts
```diff
@@ -0,0 +1,47 @@
...
+    add: false,
+    sort: false,
+    edit: false,
+    delete: false,
+  },
+  defaultView: { type: 'ViewConfig', scope: 'self' },
+  views: [],
+}
+
+export type ITableItemRow = {
```
KristianKjerstad, line 37:
```suggestion
export type TTableItemRow = {
```
prefix with T instead of I?

packages/dm-core-plugins/src/table/types.ts
```diff
@@ -0,0 +1,47 @@
+import { IUIPlugin, TViewConfig } from '@development-framework/dm-core'
+
+export type TTablePlugin = {
```
KristianKjerstad, line 3:
```suggestion
export type TTablePluginConfig = {
```

KristianKjerstad (changes requested):
Nice! A few suggestions :)

example/reset-all.sh
awesthouse: Yeah, but it stopped working when we updated dm-core-plugins version. yarn workspaces will and should install dm-core-plugins in node_modules in root folder so there was no dm-core-plugins modules in example project folder

packages/dm-core-plugins/blueprints/table/TablePluginConfig.json
KristianKjerstad: decided not necessary

packages/dm-core-plugins/blueprints/table/TableFunctionalityConfig.json
KristianKjerstad: decided not necessary

KristianKjerstad (approved):
nice!

PR #304, opened 2023-06-28
ci: change inital versions to start at 1.1.0

eoaksnes:

* Could not start at 0.1.0, since npm does not allow us to delete old packages, changing to start at 1.1.0


awesthouse (approved):
🙏🏻

PR #305, opened 2023-06-29
fix: bug with opening lists in tree

KristianKjerstad:

fix bug with opening lists in the tree. Example: orders inside DashboardExample
image

#220

### DIFF
--- a/packages/dm-core/src/domain/Tree.tsx
+++ b/packages/dm-core/src/domain/Tree.tsx
@@ -44,7 +44,13 @@ const createContainedChildren = (
     if (!attribute) return false // If no attribute, there likely where some invalid keys. Ignore those
     // Skip adding nodes for primitives
     if (!['string', 'number', 'boolean'].includes(attribute.attributeType)) {
-      const childNodeId = `${parentNode.nodeId}.${key}`
+      let childNodeId: string
+      if (Number.isInteger(Number(key))) {
+        // For arrays, bracket syntax is used to construct the node id: [key]
+        childNodeId = `${parentNode.nodeId}[${key}]`
+      } else {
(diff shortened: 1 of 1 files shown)


packages/dm-core/src/domain/Tree.tsx
```diff
@@ -44,7 +48,13 @@ const createContainedChildren = (
     if (!attribute) return false // If no attribute, there likely where some invalid keys. Ignore those
     // Skip adding nodes for primitives
     if (!['string', 'number', 'boolean'].includes(attribute.attributeType)) {
-      const childNodeId = `${parentNode.nodeId}.${key}`
+      let childNodeId: string
+      if (containsOnlyNumbers(key)) {
```
ingeridhellen, line 52: FYI: Could also have used `Number.isInteger(Number(key))`

packages/dm-core/src/domain/Tree.tsx
```diff
@@ -179,9 +189,14 @@ export class TreeNode {
...
-    this.type = Object.keys(entity).length
-      ? entity.type
-      : attribute.attributeType
+    if (Array.isArray(entity)) {
+      this.type = attribute.attributeType
+    } else {
+      this.type = Object.keys(entity).length
+        ? entity.type
+        : attribute.attributeType
+    }
```
ingeridhellen, line 198:
```suggestion
    this.type = entity && Object.hasOwn(entity, "type")
      ? entity.type
      : attribute.attributeType
```
would also work. More robust I think. If entity is an array, it will be converted to an object before testing (ie ["a", "g"] -> {"0": "a", "1": "g"}) but the keys will never be "type"

ingeridhellen (approved):
LGTM

Issue #276, opened 2023-06-26
fix: improve form-plugin


eoaksnes:
**Features:**

- [x] Be able to control what to show in the form. By default all, but should be able to specify what fields to show #355 

**Refactor:**

- [x] include types in props and function arguments @ingeridhellen

PR #306, opened 2023-06-29
refactor: add types objectfield

ingeridhellen:

Add typing to ObjectField. Also fixes a bug in the blueprint and ts types for fields

More typing makes the code more readable == easier to refactor

Refs #276

### DIFF
--- a/packages/dm-core-plugins/src/form/fields/ObjectField.tsx
+++ b/packages/dm-core-plugins/src/form/fields/ObjectField.tsx
@@ -1,24 +1,32 @@
 import {
+  EBlueprint,
   EntityPickerButton,
   EntityView,
   ErrorResponse,
   Loading,
   NewEntityButton,
+  Stack,
+  TAttribute,
+  TBlueprint,
+  TGenericObject,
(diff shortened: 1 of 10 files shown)


packages/dm-core-plugins/src/form/fields/ObjectField.tsx
```diff
@@ -51,7 +59,12 @@ const AddUncontained = (props: {
   )
 }
 
-const AddObject = (props: any) => {
+const AddObject = (props: {
+  type: string
+  namePath: string
+  onAdd: () => void
+  idReference?: string
```
KristianKjerstad, line 66: is idReference optional? 
From the usage inside the component it might seem like it's required

KristianKjerstad (approved):
Looks good!

packages/dm-core-plugins/src/form/fields/ObjectField.tsx
ingeridhellen: It is not optional but it was set that way before #290. Will fix

PR #312, opened 2023-06-29
Refactor/job plugin

KristianKjerstad:


example/src/plugins/job-ui-single/pages/JobControl.tsx
```diff
@@ -0,0 +1,74 @@
+import {
```
KristianKjerstad, line 1: content of this file is copied from JobPlugin.tsx

Issue #280, opened 2023-06-27
refactor: run a job using contained entity


eoaksnes:
- [ ] update case entity using $study.cases[<case id>].job instead of storing separate job entity
- [ ] use a reference instead of case proxy
- [ ] update job handler to use study.cases[<case id>].singal and update value
- [ ] update job API to handle references using attributes

KristianKjerstad:
pass the full id down to the jobhandler __init__.py file: the add job endpoints takes in an address, for example `DemoDataSource/$4483c9b0-d505-46c9-a157-94c79f4d7a6a.study.cases[0].job``

this should be passed down to the jobhandler

PR #315, opened 2023-06-29
refactor: job plugin in example

KristianKjerstad:

clean up the job plugin in example folder

#280


example/src/plugins/job-ui-single/pages/JobControl.tsx
```diff
@@ -0,0 +1,74 @@
+import {
```
KristianKjerstad, line 1: content of this file is copied from JobRunner.tsx

eoaksnes:
If we create a plugin for running jobs,  then this plugin should be re-usable. So all hard-coded things must be takes as input to the plugin.

KristianKjerstad:
> If we create a plugin for running jobs, then this plugin should be re-usable. So all hard-coded things must be takes as input to the plugin.

will be fixed in #382 !

PR #316, opened 2023-06-29
refactor: extract table row to own component

awesthouse:
Quick refactor that puts table row as its own component to decrease the size and categorize functionality of TablePlugin

TablePlugin component was a bit large


KristianKjerstad (approved):
Good changes!

PR #321, opened 2023-06-30
feat: ability to add custom viewrecipe in table expandable row

awesthouse:
- adds expandableRecipeViewConfig to table blueprint
- adds example expandableRecipeViewConfig to case recipe in signal app
- separate openAsTab and openAsExpandable in tableplugin so you can use both at the same time

Want to be able to add custom view in expandable that may differ from view in open/openAsTab

#258

### DIFF
--- a/packages/dm-core-plugins/blueprints/table/TablePluginConfig.json
+++ b/packages/dm-core-plugins/blueprints/table/TablePluginConfig.json
@@ -21,6 +21,11 @@
       "dimensions": "*"
     },
     {
+      "attributeType": "CORE:ViewConfig",
+      "type": "CORE:BlueprintAttribute",
+      "name": "expandableRecipeViewConfig"
+    },
+    {
       "attributeType": "./TableFunctionalityConfig.json",
       "type": "CORE:BlueprintAttribute",
       "name": "functionality",
--- a/packages/dm-core-plugins/src/table/TablePlugin.tsx
+++ b/packages/dm-core-plugins/src/table/TablePlugin.tsx
@@ -115,13 +115,15 @@ export const TablePlugin = (props: IUIPlugin) => {
       <Table style={{ width: '100%' }}>
         <Table.Head>
           <Table.Row>
-            {(config.functionality?.openAsTab ||
-              config.functionality?.openAsExpandable) && (
+            {config.functionality?.openAsExpandable && (
               <Table.Cell width="80"></Table.Cell>
             )}
             {config.columns.map((attribute: string) => (
               <Table.Cell key={attribute}>{attribute}</Table.Cell>
             ))}
+            {config.functionality?.openAsTab && (
+              <Table.Cell width="48" aria-label="Open in new tab"></Table.Cell>
+            )}
             {config.functionality?.delete && (
               <Table.Cell width="48" aria-label="Delete"></Table.Cell>
             )}
--- a/packages/dm-core-plugins/src/table/TablePlugin.tsx
+++ b/packages/dm-core-plugins/src/table/TablePlugin.tsx
@@ -116,7 +116,10 @@ export const TablePlugin = (props: IUIPlugin) => {
         <Table.Head>
           <Table.Row>
             {config.functionality?.openAsExpandable && (
-              <Table.Cell width="80"></Table.Cell>
+              <Table.Cell
+                width="80"
+                aria-label="Open as expandable"
+              ></Table.Cell>
             )}
             {config.columns.map((attribute: string) => (
               <Table.Cell key={attribute}>{attribute}</Table.Cell>
--- a/packages/dm-core-plugins/src/table/TableRow/TableRow.tsx
+++ b/packages/dm-core-plugins/src/table/TableRow/TableRow.tsx
@@ -10,11 +10,10 @@ import React, { ChangeEvent } from 'react'
 import { TTableRow } from '../types'
 import { TGenericObject, ViewCreator } from '@development-framework/dm-core'
 import {
-  add,
   chevron_down,
   chevron_up,
   delete_to_trash,
-  minimize,
+  open_in_browser,
 } from '@equinor/eds-icons'
 import { moveItem } from '../../generic-list/utils'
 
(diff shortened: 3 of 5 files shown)


packages/dm-core-plugins/src/table/TableRow/TableRow.tsx
```diff
@@ -128,6 +114,17 @@ export function TableRow(props: TTableRow) {
...
         })}
+        {config.functionality?.openAsTab && (
+          <Table.Cell style={{ textAlign: 'center' }}>
+            <Button
+              title="Open"
+              variant="ghost"
+              onClick={() => openItemAsTab(item, index)}
+            >
+              <Icon data={open_in_browser} aria-hidden /> Open
+            </Button>
```
ingeridhellen, line 125: There is several differences here relative to the Table.Cell for openAsExpanable (ie cell style, button variant, button disabled, title, aria-hidden, tooltip...). Is there a specific reason for all those differences?

packages/dm-core-plugins/src/table/TablePlugin.tsx
```diff
@@ -115,13 +115,15 @@ export const TablePlugin = (props: IUIPlugin) => {
       <Table style={{ width: '100%' }}>
         <Table.Head>
           <Table.Row>
-            {(config.functionality?.openAsTab ||
-              config.functionality?.openAsExpandable) && (
+            {config.functionality?.openAsExpandable && (
               <Table.Cell width="80"></Table.Cell>
```
ingeridhellen, line 119: Shouldn't there be an aria-label here as well?

packages/dm-core-plugins/src/table/TableRow/TableRow.tsx
```diff
@@ -178,7 +175,11 @@ export function TableRow(props: TTableRow) {
           <Table.Cell colSpan={getColumnsLength()}>
             <ViewCreator
               idReference={`${idReference}[${item.index}]`}
-              viewConfig={config.views?.[item.index] ?? config.defaultView}
+              viewConfig={
+                config.expandableRecipeViewConfig
+                  ? config.expandableRecipeViewConfig
+                  : config.views?.[item.index] ?? config.defaultView
```
ingeridhellen, line 181: I wonder if the config.views and config.defaultView makes sense anymore. I see they are defined in TTablePluginConfig, but not in TablePluginConfig.json. They don't seem to be used anywhere in our example either. What do you think @eoaksnes?

packages/dm-core-plugins/src/table/types.ts
```diff
@@ -14,6 +18,7 @@ export type TTablePluginConfig = {
   columns: string[]
   showDelete: boolean
```
ingeridhellen, line 19: EditMode and showDelete should be removed from the TTablePluginConfig, as they are no longer part of the blueprint

ingeridhellen (approved):

ingeridhellen (changes requested):
Meant to just comment, but pressed the approval button by mistake. My bad

packages/dm-core-plugins/src/table/TableRow/TableRow.tsx
ingeridhellen:
```suggestion
                  : { type: 'ViewConfig', scope: 'self' }
```

packages/dm-core-plugins/blueprints/table/TablePluginConfig.json
```diff
@@ -20,6 +20,11 @@
       "default": [],
       "dimensions": "*"
     },
+    {
+      "attributeType": "CORE:ViewConfig",
+      "type": "CORE:BlueprintAttribute",
+      "name": "expandableRecipeViewConfig"
+    },
```
ingeridhellen, line 27: Note to self: Should also add one for openAsTab at some point

packages/dm-core-plugins/src/table/TableRow/TableRow.tsx
awesthouse: disabled is definitely missing and I'll take a look, see what makes sense for each button and do some improvements

packages/dm-core-plugins/blueprints/table/TablePluginConfig.json
awesthouse: But isn't openAsTab just opening the object and should use the default view? I asked Babak and he said yes when I asked if it should just be the ui recipe for the object

packages/dm-core-plugins/src/table/types.ts
```diff
@@ -14,6 +18,7 @@ export type TTablePluginConfig = {
...
   editableColumns?: string[]
+  expandableRecipeViewConfig?: TInlineRecipeViewConfig
```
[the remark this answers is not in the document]
awesthouse: See that I've used CORE:ViewConfig in blueprint and defined type as CORE:InlineRecipeViewConfig in the ui recipe for case so there's definitely and error, but should it not be InlineRecipeViewConfig?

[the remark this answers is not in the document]
ingeridhellen: If a blueprint says that an attribute is of type MyBlueprint, and you have another blueprint MyBlueprintExtended that extends MyBlueprint, than an entity with type MyBlueprint can set that attribute to be either of type MyBlueprint or MyBlueprintExtended. So the blueprint and ui recipe is correct.
Maybe the best thing to do here would be
```suggestion
  expandableRecipeViewConfig?: TViewConfig | TInlineRecipeViewConfig | TReferenceViewConfig
```
I see that that is the viewConfig type expected by ViewCreator

packages/dm-core-plugins/blueprints/table/TablePluginConfig.json
ingeridhellen: Mh, okay. Eirik Ola thought this should be configurable. But we can discuss this on the Monday meeting. Either way, it does not need to be fixed right away

packages/dm-core-plugins/src/table/TableRow/TableRow.tsx
```diff
@@ -75,8 +75,11 @@ export function TableRow(props: TTableRow) {
       <Table.Row key={item.key}>
         {config.functionality.openAsExpandable && (
           <Table.Cell>
-            <Tooltip title={item.expanded ? 'Minimize' : 'Expand'}>
+            <Tooltip title={item.expanded ? 'Minimize' : 'Collapse'}>
```
ingeridhellen, line 78: This seems wrong. It should be Expand, right?

packages/dm-core-plugins/src/table/TableRow/TableRow.tsx
```diff
@@ -117,7 +120,7 @@ export function TableRow(props: TTableRow) {
         {config.functionality?.openAsTab && (
           <Table.Cell style={{ textAlign: 'center' }}>
             <Button
-              title="Open"
+              disabled={!item.isSaved}
```
ingeridhellen, line 123: UU is not my area of expertize, but shouldn't there be an aria-label here?

ingeridhellen (approved):

packages/dm-core-plugins/src/table/TableRow/TableRow.tsx
awesthouse: Oh, I changed the wrong one 🙈 Maybe it is time for summer holidays

packages/dm-core-plugins/src/table/TableRow/TableRow.tsx
awesthouse: It has a label/title in the button, so there's no need 👍🏻 For buttons that only have icons and no text for sr to read is where labeling it is more important

Issue #319, opened 2023-06-30
ci: improve publication of integration reports


eoaksnes:
- [x] fix: don't allow two reports to be published at the same time (set concurrency on workflow)
- [x] fix: only add a run to readme if on main branch (current check does not work) @ingeridhellen

PR #323, opened 2023-06-30
ci: handle concurrency

ingeridhellen:

- Add concurrency to on-push. Result is that if I push two times with little time in between, the first job will be cancelled when the second push is done.
- Add concurrency to the publish_report step in publish-integration-tests-report. If the job is running two at the same time, the second will have to wait until the first completes. If three runs at the same time, the first will complete, the second will be cancelled and the third will be pending. This is due to there being no way to have more than one pending job at the time https://github.com/orgs/community/discussions/5435

This issue https://github.com/equinor/dm-core-packages/actions/runs/5412414296/jobs/9836726844

Refs #319

### DIFF
--- a/.github/workflows/on-push.yaml
+++ b/.github/workflows/on-push.yaml
@@ -7,6 +7,10 @@ on:
     tags-ignore:
       - v*
 
+concurrency:
+  group: on-push-${{ github.ref_name }}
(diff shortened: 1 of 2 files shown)


KristianKjerstad (approved):
cool!

Issue #270, opened 2023-06-23
refactor: change example folder structure and files to our convention


eoaksnes:
- [x] Change folders to the convention @pbullhove 
- [x] Chage all json files to the filename.filetype.json convention (filetype is blueprint, entity, recipe .. etc) 
- [x] Add to README the convention so that. @pbullhove 

Should follow these rules:

apps: PascalCase
packages: snake_case
folders: snake_case 
blueprints: PascalCase
entities: camelCase

All blueprints should be under blueprints/
All entities should be under entities/

Files should contain what type it is:

SimpleForm.blueprint.json
SimpleForm.recipe.json -> Points to SimpleForm Blueprint

simpleFormA.entity.json
simpleFormB.entity.json
simpleFormC.entity.json

PR #328, opened 2023-06-30
Refactor/generic list to generic list

pbullhove:

Renames the generic-list plugin to generic_list according to new naming convention
This entails also changing this in all references to generic list, as well as the package name. 
This also changes the component name in the .tsx code

New naming convention

#270


KristianKjerstad (approved):
good! I like that you split the refactoring of folders into several PRs!

PR #335, opened 2023-07-03
fix: rename rest of generic list

pbullhove:

generic-list -> generic_list


ingeridhellen (commented):
Didn't we decide to remove "generic" from the name completely, @KristianKjerstad? So it would be just "list"?

ingeridhellen (approved):
LGTM!

PR #341, opened 2023-07-04
feat: improve form example

ingeridhellen:
## What does this pull request change and why?

The old "example" example was too big, and to abstract. This example removes unnecessary primitive attributes and makes the example easier to understand by using the well-known car rental company structure

Refs #276


KristianKjerstad (approved):
good! seems to work well

PR #344, opened 2023-07-05
Fix/make job in signal app work

KristianKjerstad:

fix bugs with modeling and jobhandler

[#328](https://github.com/equinor/dm-core-packages/issues/280)

### DIFF
--- a/example/job_handlers/signal-app/__init__.py
+++ b/example/job_handlers/signal-app/__init__.py
@@ -21,7 +21,8 @@ class JobHandler(JobHandlerInterface):
     def _get_by_id(self, reference: str, depth: int = 1, attribute: str = ""):
         params = {"depth": depth, "attribute": attribute}
         req = requests.get(
-            f"{config.DMSS_API}/api/documents/{reference}?resolve_links=true&depth=100", params=params, headers=self.headers  # type: ignore
+            f"{config.DMSS_API}/api/documents/{reference}?resolve_links=true&depth=100", params=params,
+            headers=self.headers  # type: ignore
         )  # type: ignore
         req.raise_for_status()
         return req.json()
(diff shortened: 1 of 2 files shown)


example/job_handlers/signal-app/__init__.py
```diff
@@ -21,30 +21,33 @@ def __init__(
...
         )  # type: ignore
         req.raise_for_status()
         return req.json()
 
     def _update(self, reference: str, document: dict):
         form_data = {k: json.dumps(v) if isinstance(v, dict) else str(v) for k, v in document.items()}
         headers = {"Authorization": f"Bearer {self.job.token}", "Access-Key": self.job.token}
-        req = requests.put(f"{config.DMSS_API}/api/documents/{reference}", data=form_data, headers=headers, params={"update_uncontained": "False"})
+        req = requests.put(f"{config.DMSS_API}/api/documents/{reference}", data=form_data, headers=headers,
+                           params={"update_uncontained": "False"})
```
ingeridhellen, line 34: What happened with the indentation?

example/job_handlers/signal-app/__init__.py
```diff
@@ -21,30 +21,33 @@ def __init__(
...
 
     def start(self) -> str:
         logger.info("Job started")
-        input_entity = self._get_by_id(f"{self.data_source}/${self.job.entity['applicationInput']['_id']}")
-        reference: str = f"{input_entity['child_id']}.signal"
-        signal_entity = self._get_by_id(reference)
-        signal_entity["value"] =  [1,2,3,4,5,6,7,8,9,10]
-        logger.info(signal_entity)
-        self._update(f"{reference}", {"data":signal_entity})
+        applicationInputReference = self.job.entity['applicationInput']['address']
```
ingeridhellen, line 40: Why change it from _id to address?

ingeridhellen (approved):
I don't know this code, so I have some questions

example/job_handlers/signal-app/__init__.py
KristianKjerstad: here is job.entity["applicationInput"] object:
```json
applicationInput: {
                        type: 'dmss://system/SIMOS/Reference',
                        referenceType: 'link',
                        address: 'dmss://DemoDataSource/$5483c9b0-d505-46c9-a157-94c79f4d7a6b'
                    },
```

with changes from https://github.com/equinor/dm-job/pull/35 we now don't resolve the job entity object by default. 
Therefore, we take the address instead of _id.

example/job_handlers/signal-app/__init__.py
KristianKjerstad: black creates a newline due to limit of line length is reached

ingeridhellen: Right, it aligns with the parenthesis. Personally, I don't like that way of indenting, since the alignment would be off if you for instance rename the `requests` variable. And since you could do the renaming automatically, you might not notice the error.

PR #346, opened 2023-07-06
fix: remove case proxy

KristianKjerstad:


KristianKjerstad:
TODO test that job running works and update branch name

PR #347, opened 2023-07-07
Fix/remove case proxy

KristianKjerstad:

Update the job example in my signal app to something that makes more sense.

- A job entity requires an applicationInput to give info on how to complete the job. I have added a flag to determine whether a long or short signal should be generated.
- use the outputTarget attribute in the job to specify where to store result from the job.
- 

#280


example/src/plugins/job-ui-single/pages/JobForm.tsx
```diff
@@ -119,12 +122,23 @@ export const JobForm = (props: {
...
+                  })
+                }}
                 value={formData[attribute.name]}
               />
             </>
           )
         })}
-        <Button onClick={() => props.onSubmit(formData)}>Submit</Button>
+        <Button
+          disabled={!formData?.applicationInput || !formData?.runner.type}
```
KristianKjerstad, line 137: disable button if user has forgotten to supply applicationInput or runner type

example/src/plugins/job-ui-single/pages/CreateJobEntity.tsx
```diff
@@ -51,6 +54,12 @@ export const CreateJobEntity = (props: TCreateJobEntityProps) => {
   const [createdJobEntity, setCreatedJobEntity] = useState<TGenericObject>()
 
   const createJobEntity = (jobEntityFormData: TJob) => {
+    if (defaultJobOutputTarget) {
+      jobEntityFormData = {
+        ...jobEntityFormData,
+        outputTarget: defaultJobOutputTarget,
```
ingeridhellen, line 60: If jobEntityFormData.outputTarget !== undefined, then it will still be overridden by the default. That seems odd.

ingeridhellen (approved):
I liked the blueprint description. We should have more of those.

example/src/plugins/job-ui-single/pages/CreateJobEntity.tsx
KristianKjerstad: agree, should only be added if jobEntityFormData does not have an outputTarget attribute

PR #355, opened 2023-07-07
feat: allow form field filtering

ingeridhellen:

Replace the order config with a fields config, that can be used both to order and to filter the form fields. If no config.fields is given, all fields are shown in the order of the blueprint

Blueprints often has a lot of meta info that does not need to be presented in a form. Being able to filter the attributes is important

Refs #276


KristianKjerstad (approved):
- added missing documentation about the config's fields attribute in Form plugin README
- fixed some variable naming and types

PR #381, opened 2023-07-31
refactor: use AttributeList directly

ingeridhellen:

Avoid using ObjectField directly from Form file

Allows for less recursiveness and stronger typing on ObjectField (will fix in new PR)

### DIFF
--- a/packages/dm-core-plugins/src/form/Form.tsx
+++ b/packages/dm-core-plugins/src/form/Form.tsx
@@ -1,11 +1,15 @@
 import * as React from 'react'
 
-import { Stack, TGenericObject } from '@development-framework/dm-core'
+import {
+  Stack,
+  TGenericObject,
+  useBlueprint,
+} from '@development-framework/dm-core'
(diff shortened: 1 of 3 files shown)


packages/dm-core-plugins/src/form/Form.tsx
```diff
@@ -43,10 +48,12 @@ export const Form = (props: TFormProps) => {
...
-            <Stack spacing={2}>
-              {type && (
-                <ObjectField config={config} namePath={namePath} type={type} />
-              )}
+            <Stack spacing={2} style={{ padding: '1rem 0' }}>
+              <AttributeList
+                namePath={namePath}
+                config={config}
+                blueprint={blueprint}
+              />
```
KristianKjerstad, line 56: I have not tested it, but I think with this change we will lose the functionality to select a custom widget in the RecipeLink.
 For example, whether to display a "textWidget" (single line text input) or a "TextareaWidget" (multi line text input)

see for example employee.recipe.json

image

KristianKjerstad (approved):

PR #384, opened 2023-08-01
refactor(form): remove superfluous control argument

ingeridhellen:

Removes the control argument to avoid unnecessary noise

control input is optional when using form provider, which we do

#276


KristianKjerstad (approved):
very good to remove this piece of unnecessary code!

PR #385, opened 2023-08-01
refactor(form): move code around

ingeridhellen:

Move files and code around. It DOES NOT change the logic in any way

- Moving Form.tsx into the component folder makes it more obvious that the FormPlugin is the starting point of the plugin
- Contexts belong in the context folder
- The attribute list is no longer merely a part of the ObjectField and should thus be in its own file

#276


packages/dm-core-plugins/src/form/components/AttributeList.tsx
```diff
@@ -0,0 +1,40 @@
+import { Stack, TAttribute, TBlueprint } from '@development-framework/dm-core'
+import React from 'react'
+import { AttributeField } from '../fields/AttributeField'
+import { TConfig } from '../types'
+
+export const AttributeList = (props: {
```
ingeridhellen, line 6: The content of this function is identical to the code block removed from ObjectField, so I wouldn't spend time reviewing it

KristianKjerstad (approved):

PR #386, opened 2023-08-01
refactor: improve TValidEntity

ingeridhellen:


packages/dm-core/src/utils/objectUtilities.ts
```diff
@@ -0,0 +1,9 @@
+export function getKey<T>(
+  object: { [k: string]: any } | undefined,
+  key: string,
+  type: string
+): T | undefined {
+  return object !== undefined && key in object && typeof object[key] == type
```
KristianKjerstad, line 6:
```suggestion
  return object !== undefined && key in object && typeof object[key] == T
```

Would this work? then we could remove the type parameter from the function.

KristianKjerstad (approved):

packages/dm-core/src/utils/objectUtilities.ts
ingeridhellen: No, it wouldn't. `T`only exist as a TS thing, it makes no sense to JS, so JS can't make logic out of it

PR #389, opened 2023-08-02
feat(form): add isInline flag

ingeridhellen:

Allows the user to select whether a contained attribute of type object should be shown inline or not.

Note: it does not work for attributes of type array or for uncontained attributes just yet. I'll do that next

image

#276

### DIFF
--- a/packages/dm-core-plugins/src/view_selector/Content.tsx
+++ b/packages/dm-core-plugins/src/view_selector/Content.tsx
@@ -1,10 +1,10 @@
-import * as React from 'react'
-import styled from 'styled-components'
 import {
   TGenericObject,
   TViewConfig,
   ViewCreator,
 } from '@development-framework/dm-core'
+import * as React from 'react'
+import styled from 'styled-components'
 import { TItemData } from './types'
 
 const HidableWrapper = styled.div<any>`
@@ -28,6 +28,7 @@ export const Content = (props: {
         <HidableWrapper
           key={config.viewId}
           hidden={config.viewId !== selectedView}
+          role="tabpanel"
         >
           <ViewCreator
             idReference={config.rootEntityId}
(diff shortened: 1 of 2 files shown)


packages/dm-core-plugins/src/view_selector/Content.tsx
```diff
@@ -28,6 +28,7 @@ export const Content = (props: {
         <HidableWrapper
           key={config.viewId}
           hidden={config.viewId !== selectedView}
+          role="tabpanel"
```
awesthouse, line 31: Nice 👍🏻 but on a sidenote - is it a bit weird that we're using css to set display: none instead of just not rendering the content the usual react way? I mean either way is fine, but we pretty much always do the latter

packages/dm-core-plugins/src/form/fields/ObjectField.tsx
```diff
@@ -174,36 +175,36 @@ export const ContainedAttribute = (props: TContentProps): JSX.Element => {
...
-              <AttributeList
-                namePath={namePath}
-                config={uiRecipe?.config}
-                blueprint={blueprint}
-              />
-            )}
-          </>
-        )}
+        {isDefined &&
+          (hasOpen && !uiAttribute?.isInline ? (
```
awesthouse, line 179: I might just be having a really hard time reading this, but should the two checks not be combined with the closing parentheses ending after isInline? 
```
(hasOpen && !uiAttribute?.isInline) ? (
  // content
)
```

awesthouse (approved):
Nice! I miiiight prefer showInline over isInline, but up to you 👍🏻

packages/dm-core-plugins/src/form/fields/ObjectField.tsx
awesthouse: So it was definitely me having a hard time reading it 😆 Doesn't mean that the code is necessarily too complex though! If I had opened it in vs code I would've had a much easier time comprehending what it said - I just wanted to double check because this kind of splitting is usually what prettier just does automatically sometimes 👍🏻

PR #393, opened 2023-08-03
Fix/cleanup file plugin

### LINKED ISSUE #358: bug: employee app example not wokring. not able to fetch image

KristianKjerstad:
clean up the file plugin
- add type TFileEntity
- fix bug with extracting data source name from idReference in FilePlugin
- fix bug in form plugin with constructing link reference.

closes #358


packages/dm-core-plugins/src/file/UploadFileButton.tsx
```diff
@@ -1,23 +1,26 @@
...
+  splitAddress,
+  TLinkReference,
+  TStorageReference,
+  useDMSS,
+} from '@development-framework/dm-core'
 import { Button, Progress } from '@equinor/eds-core-react'
 import { AxiosError } from 'axios'
 import * as React from 'react'
 import { ChangeEvent, useEffect, useRef, useState } from 'react'
-import { useRegistryContext } from '../form/context/RegistryContext'
```
KristianKjerstad, line 6: using a registry from form plugin in the file plugin does not make sense... therefore It is removed and props to component are updated

packages/dm-core/src/types.ts
```diff
@@ -132,6 +132,16 @@ export type TInlineRecipeViewConfig = TViewConfig & {
...
 }
 
+export type TFileEntity = {
+  type: string
+  name: string
+  author: string
+  date: string
+  size: number
+  filetype: string
+  content?: TLinkReference
```
ingeridhellen, line 142: According to SIMOS/File.json, content has type "binary". Is that wrong?

ingeridhellen (approved):
Nice!

packages/dm-core-plugins/src/file/UploadFileButton.tsx
ingeridhellen: Wouldn't that throw the `useRegistryContext must be used within a RegistryProvider` warning?

KristianKjerstad: it didnt seem to give any errors, but the value recieved from useRegistryContext() was sometimes empty

packages/dm-core/src/types.ts
KristianKjerstad: I was a bit confused by this. Since I don't know how the binary type is supposed to work (and I could not find any documentation about it), 

from some parts of dmss it looks like if type is "binary", it is interpreted as a storage reference.
image

However, from the implementation of the File plugin in dm-core-packages it looks like a link reference is required.
So I dont know if we should use link or storage reference for binary type.

we need to take another look at this once @eoaksnes  is back I think [(will add issue for this)](https://github.com/equinor/dm-core-packages/issues/403)

KristianKjerstad: I did try to use a storage reference instead of a link reference in the File plugin, but could not make it work

ingeridhellen: I don't know either. But I don't think it blocks this PR in any way, though. We can ask Eirik Ola next week

KristianKjerstad:
https://github.com/equinor/data-modelling-storage-service/issues/484

Issue #397, opened 2023-08-03
ci: in actions, each step should do one thing.


KristianKjerstad:
ref discussion in https://github.com/equinor/dm-core-packages/pull/388#pullrequestreview-1560836063

running reset script should be a separate step in the github action workflow 

Go throught all github actions yaml files and make sure that each step is not too big. 
- [x] codeql.yaml
- [x] create-release-pr.yaml
- [x] integration-tests.yaml
- [x] on-delete.yaml
- [x] on-main-push.yaml
- [x] on-push.yaml
- [x] on-schedule-nightly.yaml
- [x] on-tag.yaml
- [ ] publish-integration-tests-report.yaml
- [ ] remove-integration-tests-report.yaml
- [x] stale.yaml
- [x] tests.yaml

KristianKjerstad:
optimally, one bash command per step

PR #407, opened 2023-08-07
refactor(github-actions): create smaller GitHub actions steps

pbullhove:

Split up github action steps into multiple smaller steps for more readability and easier debugging. 

**Improve Readability**: By breaking down complex GitHub Actions into smaller, named steps, the code becomes easier to understand and maintain.
**Enhance Robustness**: Updating dependencies and optimizing build processes can lead to more stable and efficient builds.

This one: #397


.github/workflows/integration-tests.yaml
```diff
@@ -33,44 +33,82 @@ jobs:
...
+      - name: Pull docker images
+        run: docker-compose pull
+        working-directory: example/
+
+      - name: Capture output DMSS version
         run: |
-          docker-compose pull
           docker-compose run --rm dmss cat src/version.txt > dmss_version.txt
           dmss_version=$(tail -n 1 dmss_version.txt)
+          echo "dmss_version=$dmss_version" >> "$GITHUB_OUTPUT"
```
ingeridhellen, line 60: The step is missing id, so the variable won't be accessible. Same for the job step below. So L27 and L28 will return empty values

.github/workflows/integration-tests.yaml
```diff
@@ -33,44 +33,82 @@ jobs:
...
+        run: source .venv/bin/activate
+        working-directory: example/
+
+      - name: Install dm-cli package
+        run: pip install dm-cli
+        working-directory: example/
+
+      - name: Capture dm-cli version
+        id: capture_version
+        run: echo "version=$(dm --version)" >> "$GITHUB_OUTPUT"
```
ingeridhellen, line 111: I don't think this will work. https://stackoverflow.com/a/74669486/8231275 explains why. That's why I reactived the venv in the reset-all step below.
However, now I think the third approach makes most sense. Although I'm not sure why we would need a venv here at all. The code is already containerized.

KristianKjerstad (approved):
nice!

.github/workflows/integration-tests.yaml
```diff
@@ -33,44 +33,82 @@ jobs:
...
+      - name: Activate Python virtual environment
+        run: source .venv/bin/activate
+        working-directory: example/
+
+      - name: Install dm-cli package
+        run: pip install dm-cli
+        working-directory: example/
+
+      - name: Capture dm-cli version
+        id: capture_version
```
ingeridhellen, line 110: This id does not match the id used in L26

.github/workflows/on-main-push.yaml
```diff
@@ -40,11 +40,17 @@ jobs:
...
+
+      - name: Build packages
+        run: yarn build:packages
+
+      - name: Publish dm-core package
+        run: yarn workspace @development-framework/dm-core publish --access public
+
+      - name: Publish dm-core-plugins package
+        run: yarn workspace @development-framework/dm-core-plugins publish --access public
+
```
ingeridhellen, line 54: Both publish steps must have access to the NPM token
```suggestion
      - name: Publish dm-core package
        run: yarn workspace @development-framework/dm-core publish --access public
        env:
          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}

      - name: Publish dm-core-plugins package
        run: yarn workspace @development-framework/dm-core-plugins publish --access public
```

ingeridhellen (changes requested):
Looks a lot better!

.github/workflows/integration-tests.yaml
pbullhove: Good catch. Resolved

.github/workflows/integration-tests.yaml
pbullhove: Good catch! Fixed

.github/workflows/on-main-push.yaml
pbullhove: fixed 👍

.github/workflows/integration-tests.yaml
pbullhove: I agree that we don't need a venv in github actions. 
I removed the setup of the venv.

ingeridhellen (approved):

PR #411, opened 2023-08-08
feat(form): allow inline presentation of list of non-primitives

ingeridhellen:

Earlier, you could display a list of any attribute, both complex and primitive, but the former would not look good. This PR fixes that by calling EntityView for list of complex values, allowing them to be shown as defined by their uiRecipe

image

#276


e2e/tests/plugin-form-Nested.spec.ts
```diff
@@ -93,28 +93,47 @@ test('Adding a trainee', async () => {
...
+})
+
 test('New car', async () => {
-  await page.getByText('CarsOpen').getByRole('button', { name: 'Open' }).click()
-  await expect.soft(page.getByText('1 - 2 of 2')).toBeVisible()
-  await page.getByRole('button', { name: 'Append Add Item' }).click()
-  await expect.soft(page.getByText('1 - 3 of 3')).toBeVisible()
-  await page.getByRole('button', { name: 'Save' }).click()
-  await page.getByRole('button', { name: 'Open item' }).last().click()
+  const cars = page.getByTestId('cars')
```
KristianKjerstad, line 118:
```suggestion
  const carsPageElement = page.getByTestId('cars')
```

maybe rename the variable to something a bit more descriptive. Just an example name provided.

ingeridhellen: I went for carsDiv - short and descriptive. I also renamed the locations variable in the same manner.

PR #413, opened 2023-08-09
refactor(entityPicker): update return value

ingeridhellen:

Changes the input values of EntityPicker

Getting both the address and the entity can be nice, for instance if you want to insert a reference (ie need address), but also want to ensure that the entity type is correct.

### DIFF
--- a/example/src/plugins/job-ui-single/pages/JobForm.tsx
+++ b/example/src/plugins/job-ui-single/pages/JobForm.tsx
@@ -1,14 +1,14 @@
 import {
   EBlueprint,
-  TJob,
-  Stack,
-  useBlueprint,
-  Loading,
-  TAttribute,
   EntityPickerButton,
   JobStatus,
-  TGenericObject,
+  Loading,
+  Stack,
+  TAttribute,
+  TJob,
   TLinkReference,
+  TValidEntity,
+  useBlueprint,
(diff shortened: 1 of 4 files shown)


example/src/plugins/job-ui-single/pages/JobForm.tsx
```diff
@@ -98,11 +97,15 @@ export const JobForm = (props: {
                   </p>
                 </div>
                 <EntityPickerButton
-                  returnLinkReference={true}
-                  onChange={(linkReferenceEntity: TLinkReference) => {
+                  onChange={(address: string, entity: TValidEntity) => {
+                    const ref: TLinkReference = {
```
KristianKjerstad, line 101: I generally prefer full names over abbreviations :) 

```suggestion
                    const linkReference: TLinkReference = {
```

KristianKjerstad (approved):

example/src/plugins/job-ui-single/pages/JobForm.tsx
ingeridhellen: I generally agree, though I find abbreviations useful when the scope is small, ie `[1,2,3].map((x) => x*x)`. But this scope can easily be extended, and then an abbreviation would be bad. Will fix

PR #428, opened 2023-08-16
feat: view-selector example

roar-larsen:
Added new example case for the view selector. The example uses both sidebar and tabs.

For demonstration and testing purposes.

### DIFF
--- a/example/app/data/DemoDataSource/plugins/view_selector/car_garage/blueprints/Owner.blueprint.json
+++ b/example/app/data/DemoDataSource/plugins/view_selector/car_garage/blueprints/Owner.blueprint.json
@@ -1,0 +1,28 @@
+{
+  "type": "CORE:Blueprint",
+  "name": "OwnerBlueprint",
+  "extends": ["CORE:NamedEntity"],
+  "attributes": [
+    {
+      "attributeType": "string",
+      "type": "CORE:BlueprintAttribute",
+      "name": "nameOwner",
+      "label": "Name of Owner",
+      "optional": false
+    },
+    {
+      "attributeType": "string",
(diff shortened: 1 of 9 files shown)


roar-larsen:
Integration tests are failing because of an error in DMSS.

example/app/data/DemoDataSource/plugins/view_selector/car_garage/blueprints/Owner.blueprint.json
```diff
@@ -0,0 +1,28 @@
...
+  "name": "OwnerBlueprint",
+  "extends": ["CORE:NamedEntity"],
+  "attributes": [
+    {
+      "attributeType": "string",
+      "type": "CORE:BlueprintAttribute",
+      "name": "name",
+      "label": "Name of Owner",
+      "optional": false
+    },
```
KristianKjerstad, line 12:
```suggestion

```

name in the attributes list is not needed since it is included in NamedEntity which OwnerBlueprint extends from

KristianKjerstad (approved):

example/app/data/DemoDataSource/plugins/view_selector/car_garage/blueprints/Owner.blueprint.json
roar-larsen: But then I can´t change the label, can I?

KristianKjerstad: oh right, that is not possible. 
The name can  be kept in the attributes list then 👍

PR #431, opened 2023-08-17
chore: update pre-commit

ingeridhellen:

- Updated pre-commit hook versions
- Add pretty-format-json hook and stop prettier from fixing json files
- Add end-of-file-fixer hook

Note: also tried adding codespell, but that gave to many false positives for me to find valuable

The pretty-format-json hook allows for sorting keys. Having the name key always on top makes the blueprints easier to read in my opinion.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -21,7 +21,10 @@
     "ecmaVersion": "latest",
     "sourceType": "module"
   },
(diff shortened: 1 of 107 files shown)


KristianKjerstad (approved):
nice, this is very useful

PR #433, opened 2023-08-22
fix(form): use documentAdd instead of Update

ingeridhellen:

- Remove workarounds in tests caused by #258
- Use documentAdd both when selecting reference and adding object

The CI was failing


pbullhove (approved):
Approved after sync PR

PR #439, opened 2023-08-24
refactor: remove button, header and clipboard copy

ingeridhellen:

- Removes button component and switch to eds instead
- Removes unused header component
- Remove react-copy-to-clipboard, since we don't really need a library to do that


eoaksnes (approved):
Nice cleanup

PR #440, opened 2023-08-24
test: playwright test for view-selector example

roar-larsen:
Added Playwright test for the view-selector example.


eoaksnes (approved):
Nice

PR #446, opened 2023-08-25
fix(job): use check document existence endpoint to verify job

### LINKED ISSUE #437: fix(job): can't start signalApp job
#### Actual Behaviour

Raises exception 400 bad request. 

#### Steps to Reproduce

- Open the example application
- Go to signalApp
- Open case1
- Push the create button for the job

collinlokken:
Introduces a check that verifies the existence of the job before it is attempted updated. The documentAdd endpoint will be used instead if the job does not exist.

The check removes an error in the client, produced by the fact that a non-existent job is attempted updated.

Closes #437

### DIFF
--- a/example/src/plugins/job-ui-single/pages/CreateJobEntity.tsx
+++ b/example/src/plugins/job-ui-single/pages/CreateJobEntity.tsx
@@ -41,9 +41,6 @@ export const CreateJobEntity = (props: TCreateJobEntityProps) => {
   } = props
 
   const DmssApi = useDMSS()
-  const destinationIsAPackage: boolean = !(
-    jobEntityDestination.includes('.') || jobEntityDestination.includes('[')
-  )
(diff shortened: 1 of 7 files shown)


example/src/plugins/job-ui-single/pages/CreateJobEntity.tsx
```diff
@@ -54,31 +51,63 @@ export const CreateJobEntity = (props: TCreateJobEntityProps) => {
...
+      if (jobExists) {
+        updateDocument(jobEntityDestination, jobEntityFormData)
+      } else {
+        addDocument(jobEntityDestination, jobEntityFormData)
+      }
+    }
+
+    // TODO: Implement document check
+    DmssApi.documentCheck({
+      address: jobEntityDestination,
```
ingeridhellen, line 103: Can the address point to a list? In that case, we should use addDocument even if the list exist, I think

ingeridhellen (approved):

PR #481, opened 2023-09-05
feat: view only mode for form plugin

### LINKED ISSUE #463: feat: create a readonly flag for form, so that it becomes a viewer
So the whole form is in edit mode. 

Optionally, we can have 
- edit mode or readonly mode so taht

collinlokken:
View only feature for form plugin that gives the opportunity to configure a "readOnly" attribute in the recipe. This makes a button appear that allows toggling between edit and readOnly mode

Feature request

Closes  #463


pbullhove (approved):

pbullhove:
💥Boom. Nice work bro

packages/dm-core-plugins/src/form/components/AttributeList.tsx
```diff
@@ -35,10 +43,18 @@ export const AttributeList = (props: {
...
+          readOnly={readOnly}
         />
       </div>
     )
   })
 
-  return <Stack spacing={1}>{attributeFields}</Stack>
+  return (
+    <Stack spacing={1}>
+      {config?.readOnly && (
```
ingeridhellen, line 54: This way, if the config is set to readonly, it will always be possible to edit the form anyway be toggling. Thus it is still impossible to create a form where editing is impossible, which I think is what Babak wanted.

I think that if we want a toggle button or not should be defined by yet another config setting. And then perhaps rename config.readonly to config.initiallyReadOnly.

collinlokken: Agreed!

PR #482, opened 2023-09-06
test: signal app create job

roar-larsen:
Playwright test for create job in SignalApp.

There are 2 snapshot images used as reference image. One is for the CI and the other is there for running locally.


ingeridhellen (approved):
:tada

PR #485, opened 2023-09-07
test(form): add example of storage uncontained

ingeridhellen:

Adds simple example

Need to make an example to work with when adapting form to storage uncontained.


pbullhove (approved):
Very nice. Clean, simple and understandable

Issue #469, opened 2023-08-30
fix(table): can't remove 0 in number field


ingeridhellen:
- [ ] Open signalApp -> Open case1 -> Open Components -> select a number field and try removing the text there, ie try replacing 0 with 1. You won't be able to have an empty string there.

PR #489, opened 2023-09-08
Makes it possible to enter 1 without a padded 0 in table row

henrikbossart:
Changes `TableRow.tsx` to allow users to enter single digit numbers without left-padded 0 to be added.

It fixed the issue described above.

#469


ingeridhellen (approved):
Not what I had in mind, but better! A bit hard to test it properly, as it seems the table input doesn't handle me writing "0." either, but we can fix that afterwards. Nice job!

henrikbossart:
Hmm... I guess we could either debounce the event and strip the trailing "." after X seconds, or change the event to `onBlur` and move the formatting logic to the `updateItem` function instead. Not sure which (if any) of those approaches fits best with the intended use of the table.

ingeridhellen:
Third alternative would be to not do anything until the user clicks save, then do validation and parsing then. I think you can merge what you got, and then consider the rest later on

PR #490, opened 2023-09-08
fix(form): ObjectField use default value if defined

### LINKED ISSUE #470: fix(form): if complex, non-optional attribute has default value, use that instead of instantiateEntity
Also, for primitive, optional values the default value should in some way be reflected in the form (parent default supersedes the child default)

collinlokken:
If a complex attribute (Object) has not been instantiated, but has a default value defined, the default value should be used upon creation. 

The instantiateEntity endpoint is currently used if we want to add a complex attribute, but it should instead first check if a default value is available in the blueprint.

Closes #470

### DIFF
--- a/packages/dm-core-plugins/src/form/fields/ObjectField.tsx
+++ b/packages/dm-core-plugins/src/form/fields/ObjectField.tsx
@@ -74,35 +74,43 @@ const SelectReference = (props: { type: string; namePath: string }) => {
   )
 }
 
-const AddObject = (props: { type: string; namePath: string }) => {
-  const { type, namePath } = props
+const AddObject = (props: {
+  type: string
+  namePath: string
+  defaultValue: any
+}) => {
+  const { type, namePath, defaultValue } = props
   const { setValue } = useFormContext()
   const dmssAPI = useDMSS()
   const { idReference } = useRegistryContext()
   const handleAdd = () => {
+    if (!defaultValue) {
+      dmssAPI
+        .instantiateEntity({
+          entity: { type: type as string },
+        })
+        .then((newEntity: AxiosResponse<any>) => addDocument(newEntity.data))
+    } else {
+      addDocument(defaultValue)
+    }
+  }
+  const addDocument = (document: any) => {
     const options = {
       shouldValidate: true,
       shouldDirty: true,
       shouldTouch: true,
     }
-
     dmssAPI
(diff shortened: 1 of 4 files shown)


packages/dm-core-plugins/src/form/fields/ObjectField.tsx
```diff
@@ -85,24 +89,38 @@ const AddObject = (props: { type: string; namePath: string }) => {
...
+  }
+  const addDocument = (
+    document: any,
+    options: {
+      shouldValidate: boolean
+      shouldDirty: boolean
+      shouldTouch: boolean
+    }
+  ) => {
+    console.log(document)
```
ingeridhellen, line 112:
```suggestion
```

packages/dm-core-plugins/src/form/fields/ObjectField.tsx
```diff
@@ -85,24 +89,38 @@ const AddObject = (props: { type: string; namePath: string }) => {
...
       shouldTouch: true,
     }
-
+    if (!defaultValue) {
+      dmssAPI
+        .instantiateEntity({
+          entity: { type: type as string },
+        })
+        .then((newEntity: AxiosResponse<any>) =>
+          addDocument(newEntity.data, options)
```
ingeridhellen, line 98: If you moved the `options` object on L87 out of handleAdd and into addDocument, you wouldn't have to pass it into AddDocument, and then you wouldn't have to add a type for it either. Also, the object would be declared closer to where it is used

ingeridhellen (changes requested):

packages/dm-core-plugins/src/form/components/AttributeList.tsx
```diff
@@ -44,6 +44,7 @@ export const AttributeList = (props: {
           attribute={attribute}
           uiAttribute={uiAttribute}
           readOnly={readOnly}
+          defaultValue={attribute.default}
```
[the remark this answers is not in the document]
collinlokken: very true

packages/dm-core-plugins/src/form/fields/ObjectField.tsx
collinlokken: Yes, very true!

ingeridhellen (approved):
Looks good!

PR #494, opened 2023-09-11
refactor: remove golden-layout

### LINKED ISSUE #491: refactor: remove golden layout

henrikbossart:
Removes the `golden-layout` package from the app, and replaces its usage with `EntityView`.

Layout functionality was broken, and unused.

closes #491


ingeridhellen (approved):
Nice!

I notice we're using a ModalProvider, that perhaps made sense when we used golden-layout. But perhaps we can remove it now?

PR #499, opened 2023-09-12
fix: remove all edit buttons when in readOnly mode

### LINKED ISSUE #496: fix: disable all form edit buttions when in read only mode

collinlokken:
Removes the [Add and Save], [Select and Save], [Add], [Remove] and [Add Item] (ListPlugin) buttons when in read only mode.

It should not be possible to delete or add objects when in read only mode.

Closes #496

### DIFF
--- a/example/app/data/DemoDataSource/recipes/plugins/form/read_only/customer.recipe.json
+++ b/example/app/data/DemoDataSource/recipes/plugins/form/read_only/customer.recipe.json
@@ -4,7 +4,11 @@
   "initialUiRecipe": {
     "name": "form",
     "type": "CORE:UiRecipe",
-    "plugin": "@development-framework/dm-core-plugins/form"
+    "plugin": "@development-framework/dm-core-plugins/form",
+    "config": {
+      "type": "PLUGINS:dm-core-plugins/form/FormInput",
+      "readOnly": true
+    }
   },
   "uiRecipes": [
     {
--- a/packages/dm-core-plugins/src/form/components/AttributeList.tsx
+++ b/packages/dm-core-plugins/src/form/components/AttributeList.tsx
@@ -1,8 +1,7 @@
-import { Stack, TAttribute, TBlueprint } from '@development-framework/dm-core'
-import React, { useState } from 'react'
+import { TAttribute, TBlueprint } from '@development-framework/dm-core'
+import React from 'react'
 import { AttributeField } from '../fields/AttributeField'
 import { TConfig } from '../types'
-import { Button } from '@equinor/eds-core-react'
 
 export const AttributeList = (props: {
   namePath: string
@@ -10,14 +9,6 @@ export const AttributeList = (props: {
   blueprint: TBlueprint | undefined
 }) => {
   const { namePath, config, blueprint } = props
-
-  const [readOnly, setReadOnly] = useState<boolean | undefined>(
-    config?.readOnly
-  )
-  const toggleHandler = () => {
-    setReadOnly(!readOnly)
-  }
-
   const prefix = namePath === '' ? `` : `${namePath}.`
 
   const attributes = blueprint?.attributes ?? []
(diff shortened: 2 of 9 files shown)


packages/dm-core-plugins/blueprints/list/ListPluginConfig.json
```diff
@@ -52,6 +52,13 @@
       "dimensions": "*",
       "default": [],
       "optional": false
+    },
+    {
+      "name": "readOnly",
```
ingeridhellen, line 57: This is not needed. Whether or not list is readonly should be determined by the config.functionality. It isn't currently, because list is temporarily deprecated and no one works on it, but that is the idea of it at least.

Perhaps you could use the table plugin instead?

example/app/data/DemoDataSource/recipes/plugins/form/read_only/customer.recipe.json
```diff
@@ -18,6 +22,27 @@
...
+        "headers": [
+          "name",
+          "plateNumber"
+        ],
+        "openInline": false,
+        "functionality": {
+          "type": "PLUGINS:dm-core-plugins/list/FunctionalityConfig",
+          "delete": true
+        },
+        "readOnly": true
```
ingeridhellen, line 44: Please remove this

packages/dm-core-plugins/src/list/ListPlugin.tsx
```diff
@@ -284,14 +287,16 @@ export const ListPlugin = (props: IUIPlugin & { config?: TListConfig }) => {
...
         />
-        <Stack direction="row" spacing={1}>
-          <AppendButton onClick={() => addItem()} />
-          <SaveButton
-            onClick={save}
-            disabled={!unsavedChanges}
-            isLoading={isLoading}
-          />
-        </Stack>
+        {!internalConfig.readOnly && (
```
ingeridhellen, line 290: If you need to use list with read-only, please instead use `internalConfig.functionality.add` and wrap the append button in it. For the save button you could check if both functionality.delete, .add, .edit and .sort are false, if so hide the save button

packages/dm-core-plugins/src/form/components/AttributeList.tsx
```diff
@@ -43,18 +34,11 @@ export const AttributeList = (props: {
...
 
-  return (
-    <Stack spacing={1}>
-      {config?.editToggle && (
-        <Button onClick={toggleHandler}>{readOnly ? 'Edit' : 'View'}</Button>
-      )}
-      {attributeFields}
-    </Stack>
-  )
+  return <Stack spacing={1}>{attributeFields}</Stack>
```
ingeridhellen, line 43: You don't really need the `Stack` either
```suggestion
  return <>{attributeFields}</>
```

packages/dm-core-plugins/src/form/fields/ObjectField.tsx
```diff
@@ -308,6 +319,7 @@ export const ObjectField = (props: TObjectFieldProps): JSX.Element => {
       label={displayLabel}
       type={type === 'object' && values ? values.type : type}
       defaultValue={defaultValue}
+      readOnly={readOnly}
```
ingeridhellen, line 322: readOnly is already passed on L317, so there is no need for this nor the unwrapping of readonly from the props on L303

ingeridhellen (changes requested):
The hiding of buttons in ObjectField looks good, but this PR is trying to do too much things that don't really relate. Removing editToggle causes issues for Roar's #501, so it is unfortunate that such a major change is hidden inside this PR. And adding a config to list is also something that deserves its own PR and more attention.

example/app/data/DemoDataSource/recipes/plugins/form/read_only/car.recipe.json
```diff
@@ -17,5 +17,28 @@
       ],
       "readOnly": true
     }
-  }
+  },
+  "uiRecipes": [
+    {
+      "name": "MyCustomList",
```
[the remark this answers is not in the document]
collinlokken: removed

example/app/data/DemoDataSource/recipes/plugins/form/read_only/customer.recipe.json
collinlokken: removed 👍

packages/dm-core-plugins/src/list/ListPlugin.tsx
collinlokken: Landed on not addressing this in this PR

ingeridhellen (approved):
Nice!

PR #500, opened 2023-09-12
Fix/minor UI changes

henrikbossart:
Made some minor adjustments to a couple of components, nothing big.
Most notable change is the removal of the `RecipeSelector` component in favor of the EDS `Menu` component.

Other than that there is a bit more padding given to the YAML, a slight change to the height of the app, and a change in wording on a button.


example/src/App.tsx
```diff
@@ -19,12 +19,13 @@ function App() {
       style={{
         display: 'flex',
         padding: '20px',
-        overflow: 'auto',
+          height: '100vh'
```
ingeridhellen, line 22: The indentation seems off

packages/dm-core-plugins/src/header/HeaderPlugin.tsx
```diff
@@ -28,7 +27,10 @@ const Icons = styled.div`
   }
 `
 
-const ClickableIcon = styled.div`
+const ClickableIcon = styled.button`
```
ingeridhellen, line 30: I think this component if pre-EDS. The styling is wrong and it lacks a aria-label. We should instead use something like the TooltipButton I made.

packages/dm-core-plugins/src/header/HeaderPlugin.tsx
```diff
@@ -99,31 +102,42 @@ export default (props: IUIPlugin): JSX.Element => {
...
-                setSelectedRecipe(getRecipeConfigAndPlugin(uiRecipeName))
-              }}
-            />
-          )}
+          <Menu open={appSelectorOpen} anchorEl={anchorEl}>
+            {recipeNames.map((recipe, index) => (
+              <Menu.Item
+                key={index}
+                onClick={() =>
+                  setSelectedRecipe(getRecipeConfigAndPlugin(recipe))
```
ingeridhellen, line 133:
```suggestion
                  setSelectedRecipe(getRecipeConfigAndPlugin(recipe))
                  setAppSelectorOpen(false)
```
would be even better

packages/dm-core-plugins/src/header/HeaderPlugin.tsx
```diff
@@ -99,31 +102,42 @@ export default (props: IUIPlugin): JSX.Element => {
...
+                }
+              >
+                {recipe}
+              </Menu.Item>
+            ))}
+          </Menu>
+          <h4 style={{ paddingLeft: 10 }}>{entity.label}</h4>
```
ingeridhellen, line 140: We should try to replace every header element we find with EDS Typography

ingeridhellen (approved):
Nice change! Happy to see the RecipeSelector go.

Since the project is older than EDS, we have a lot of custom features that could have been solved better with EDS - like the RecipeSelector.

example/src/App.tsx
henrikbossart: indeed... somehow my editor doesn't detect the config, and thus doesn't auto lint on save...... incredibly frustrating

packages/dm-core-plugins/src/header/HeaderPlugin.tsx
henrikbossart: indeed!

packages/dm-core-plugins/src/header/HeaderPlugin.tsx
henrikbossart: I suspect there are a couple of headings, so I'll open a new PR for that 👍

henrikbossart:
FINALLY!

Issue #459, opened 2023-08-30
feat(UI): Modify UI job runner plugin, by adding Start Arrow and Stop icons.


pbullhove:
For now the signal job ui is quite messy. Want to have it very _user_-friendly, not for a developer, but a user of the application

- [x] With start arrow > and stop button. 
    - [x]  Find nice EDS iconds for this. 
- [x] Have "run again" job after the job is finished. 
- [ ] Either manual-refresh button or continously ping every 2/5 seconds while the job is running.

henrikbossart:
Am I correct in assuming that this signal app doesn't actually work atm? and that the data generated when running the simulation is just a hard coded random example? 🤔

PR #502, opened 2023-09-13
Feat/job UI

henrikbossart:
Just a few tweaks to the Job runner UI. I intended to make it possible to ping the api for getting updates on the job, but it seems to me that the job isn't actually running properly? Seems more like a hard coded random example.

I guess the actual logic of re-running and pinging the jobs can be a different PR though 🤷‍♂️ 

#459


example/src/plugins/job-ui-single/pages/JobControl.tsx
```diff
@@ -21,58 +28,59 @@ export const JobControl = (props: { jobEntityId: string }) => {
...
-        <Button
-          onClick={() => {
-            start()
-            setJobIsStarted(true)
-          }}
-          variant="contained"
+      </JobButtonWrapper>
+      {status === JobStatus.Failed && (
+        <Card
+          variant={status === JobStatus.Failed ? 'danger' : 'info'}
```
ingeridhellen, line 64: This will always evaluate to 'danger'

packages/dm-core/src/hooks/useJob.tsx
```diff
@@ -114,22 +117,30 @@ export function useJob(entityId?: string, jobId?: string): IUseJob {
...
       })
+      setStatus(JobStatus.Failed)
       return null
     }
     setIsLoading(true)
+    setStatus(JobStatus.Starting)
     return dmJobApi
       .startJob({ jobDmssId: entityId })
       .then((response: AxiosResponse<StartJobResponse>) => {
+        console.log(response)
```
ingeridhellen, line 128:
```suggestion
```

ingeridhellen (approved):
I don't know this code very well, but LGTM

example/src/plugins/job-ui-single/pages/JobControl.tsx
henrikbossart: jupp, forgot to change

PR #506, opened 2023-09-14
refactor: remove entityPickerInput

ingeridhellen:

With the extraction of button from EntityPickerButton, this component can easily be replaced by an Input and a EntityPickerDialog


henrikbossart (approved):
Doesn't seem to be used anywhere anyway, so I guess this is fine 🤷‍♂️

PR #507, opened 2023-09-14
use splitAddress to split addresses and id references

### LINKED ISSUE #396: refactor: use the dm-core utility function splitAddress() to split id references in all pluings
right now, several different ways are used to split the idReference. Example:
JobControlplugin.tsx

Image

ObjectField:

Image

We should use the splitAddress utility function everywhere in the code base

henrikbossart:
changes how addresses and id references are split.

closes #396


packages/dm-core-plugins/src/form/fields/BinaryField.tsx
```diff
@@ -14,9 +15,9 @@ const getTarget = (initialValue: any) => {
   const { idReference } = useRegistryContext()
   if (initialValue['address'].includes('blob')) {
     const address = initialValue['address'].replace('blob://', '')
-    return address.split('/', 2)
+    return splitAddress(address).dataSource
```
ingeridhellen, line 18: getTarget should return an array of length 2.

ingeridhellen (changes requested):

packages/dm-core-plugins/src/form/fields/BinaryField.tsx
henrikbossart: right 😶 🤦

ingeridhellen (approved):

PR #508, opened 2023-09-15
add slight padding to the content and top bar.

### LINKED ISSUE #498: Fix: Adjust the padding on EntityView
There is little to no padding in nested instances of the `EntityView` component, which makes the interfaces kinda cramped.

henrikbossart:
Adds slight padding to the content of views. Not quite sure if its actually better or not, but the EDS states that minimum distance between components should be 8px, so..

closes #498


packages/dm-core-plugins/src/view_selector/ViewSelectorPlugin.tsx
```diff
@@ -147,6 +147,12 @@ export const ViewSelectorPlugin = (
         />
       )}
       <Content
+        style={{
+          ...(internalConfig.asSidebar
+            ? { paddingLeft: '8px' }
+            : { paddingTop: '8px' }),
+          paddingRight: '8px',
+        }}
```
ingeridhellen, line 155: Why pass this? Couldn't you achieve the same by making a styled-component with a prop out of Content? Or wrapping Content in a div with style?

ingeridhellen (approved):

packages/dm-core-plugins/src/view_selector/ViewSelectorPlugin.tsx
henrikbossart: Force of habit. Not used to the concept of styled components approach.

> Or wrapping Content in a div with style?

I'm just more used to this approach (although not having to take in styles as a prop...)

Issue #488, opened 2023-09-07
fix(form): improve button layout to make it more compact


ingeridhellen:
As requested by Babak. Also use the vertical line more

PR #509, opened 2023-09-15
fix(form): use icons for form buttons

ingeridhellen:

Form uses icon buttons with tooltip for opening, adding, editing and removing nested objects and list.
Also created a generic TooltipButton and adapted tests to the new layout

image

ref #488


packages/dm-core-plugins/src/common/TooltipButton.tsx
henrikbossart, line 1: Why not take in the icon name/object as a prop in this component and render it if provided, instead of importing the Icon component in a ton of views?

henrikbossart (approved):
👌

PR #513, opened 2023-09-18
refactor: list example

roar-larsen:
Added "row" as a role to the list plugin.
Updated the Playwright test to utilise the changes.

Test are easier and much prettier.


ingeridhellen (approved):
🎉

PR #517, opened 2023-09-19
refactor(explorer): extract dialogs

ingeridhellen:

The ContextMenu file was long and hard to get a grasp of, so I split it into multiple files. Also noticed that we were rendering 6 dialogs for each node in the tree, which seems a bit excessive, so I added an inline if around each render

### DIFF
--- a/packages/dm-core-plugins/src/explorer/components/context-menu/NodeRightClickMenu.tsx
+++ b/packages/dm-core-plugins/src/explorer/components/context-menu/NodeRightClickMenu.tsx
@@ -12,7 +12,7 @@ import { createContextMenuItems } from './utils/createContextMenuItmes'
 export const STANDARD_DIALOG_WIDTH = '100%'
 export const STANDARD_DIALOG_HEIGHT = '300px'
 
-export const NodeRightClickMenu = (props: TNodeWrapperProps) => {
+const NodeRightClickMenu = (props: TNodeWrapperProps) => {
   const { node, children } = props
   const [dialogId, setDialogId] = useState<string>('')
   const [formData, setFormData] = useState<any>('')
@@ -105,3 +105,5 @@ export const NodeRightClickMenu = (props: TNodeWrapperProps) => {
     </>
   )
 }
+
+export default NodeRightClickMenu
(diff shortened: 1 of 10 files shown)


packages/dm-core-plugins/src/explorer/components/context-menu/NodeRightClickMenu.tsx
```diff
@@ -0,0 +1,109 @@
...
+        open={showMenu}
+        onClose={() => setShowMenu(false)}
+        anchorEl={anchorEl}
+        placement="bottom-start"
+        matchAnchorWidth={true}
+      >
+        {menuItems}
+      </Menu>
+
+      {dialogId === 'new-folder' && (
```
eoaksnes, line 48: This is old stuff, but maybe would be nice to use some enum or something for the dialog ids?

packages/dm-core-plugins/src/explorer/components/dialogs/DeleteDialog.tsx
```diff
@@ -0,0 +1,54 @@
...
+      </Dialog.Header>
+      <Dialog.CustomContent>
+        Are you sure you want to delete the entity <b>{node.name}</b> of type{' '}
+        <b>{node.type}</b>?
+      </Dialog.CustomContent>
+      <Dialog.Actions>
+        <Button
+          color="danger"
+          onClick={async () => {
+            await DeleteAction(node, dmssAPI, setLoading)
```
eoaksnes, line 39: This is old stuff, but does it make sense to have the action (DeleteAction) in a different file? It is only used inside this file, right?

eoaksnes (approved):
Very nice cleanup!

packages/dm-core-plugins/src/explorer/components/dialogs/DeleteDialog.tsx
ingeridhellen: Fix in new PR

packages/dm-core-plugins/src/explorer/components/context-menu/NodeRightClickMenu.tsx
ingeridhellen: Will fix in new PR

PR #520, opened 2023-09-19
refactor(yaml): remove style.css

ingeridhellen:

Removes the css file, thus avoiding the extra copy step in the building. Looks mostly the same though, except some changes in padding: 
image


pbullhove (approved):
Nice!

Issue #521, opened 2023-09-19
refactor(explorer): extract dialogs


ingeridhellen:
#### Describe Problem

The dialogs are all defined in the same file, making it large and cluttered

#### Suggest Changes

Split into multiple files and refactor the code

#### Provide Examples

PR #522, opened 2023-09-19
refactor(explorer): create Dialog enum

ingeridhellen:

Introduces a EDialog enum that represents the dialog ids

Avoid bugs caused by spelling mistakes ++

ref #521

### DIFF
--- a/packages/dm-core-plugins/src/explorer/components/context-menu/NodeRightClickMenu.tsx
+++ b/packages/dm-core-plugins/src/explorer/components/context-menu/NodeRightClickMenu.tsx
@@ -1,6 +1,7 @@
 import { TNodeWrapperProps } from '@development-framework/dm-core'
 import { Menu } from '@equinor/eds-core-react'
 import React, { useState } from 'react'
+import { EDialog } from '../../types'
 import AppendEntityDialog from '../dialogs/AppendEntityDialog'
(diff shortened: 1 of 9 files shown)


packages/dm-core-plugins/src/explorer/components/context-menu/NodeRightClickMenu.tsx
```diff
@@ -1,6 +1,7 @@
 import { TNodeWrapperProps } from '@development-framework/dm-core'
 import { Menu } from '@equinor/eds-core-react'
 import React, { useState } from 'react'
+import { EDialog } from '../../types'
```
henrikbossart, line 4: Should we maybe change the imports to be absolute rather than relative? 🤔

henrikbossart (approved):

packages/dm-core-plugins/src/explorer/components/context-menu/NodeRightClickMenu.tsx
henrikbossart: Or, I mean, not something _to be fixed_ by this PR 😅

ingeridhellen: Mh. My editor just auto-generate this for me, so I haven't given it much thought. I suppose the readability would be better, though

PR #525, opened 2023-09-20
refactor(explorer): move actions into dialogs

ingeridhellen:

Too much, really. The onClick handlers were organized in different ways for each dialog, some in an action file, some inline. I've changed this so that all the event handlers are defined in the component that use them, but outside the render tree. 
I also noticed while working with this that the formData and loading states were only used inside each dialog, so it really made no sense to pass them down from the parent (my bad). Fixed that, and renamed the formData state to something more descriptive.
Also renamed createContextMenuItmes to MenuItems and made it a component.
Finally, I realized that the root package dialog was basically identical to the folder dialog, so I made some small changes to the latter and deleted the former.

#521


packages/dm-core-plugins/src/explorer/components/context-menu/NodeRightClickMenu.tsx
henrikbossart, line 1: When right clicking an element in the tree, and try to click one of the items, you have to click twice for it to register the click.

This is probably not because of this PR, but its an issue nonetheless.

packages/dm-core-plugins/src/explorer/components/context-menu/NodeRightClickMenu.tsx
```diff
@@ -7,22 +7,17 @@ import DeleteDialog from '../dialogs/DeleteDialog'
...
   const { node, children } = props
   const [dialogId, setDialogId] = useState<EDialog | undefined>()
-  const [formData, setFormData] = useState<any>('')
-  const [loading, setLoading] = useState<boolean>(false)
   const [showMenu, setShowMenu] = useState<boolean>(false)
   const [anchorEl, setAnchorEl] = useState<HTMLElement | null>(null)
 
-  const menuItems = createContextMenuItems(node, setDialogId)
-
   //TODO when the tree changes by adding new package or deleting something, the tree should be updated to give consistent UI to user
```
henrikbossart, line 21: This todo seems to be done already? 🤔

henrikbossart (approved):
This seems fine, just two small comments, see above ☝️

PR #528, opened 2023-09-21
test: make uncontained form test more robust

ingeridhellen:

Most importantly it makes the test more robust by ensuring that the page is done loading before playwright moves on. Also refactored a bit

The test was failing locally


roar-larsen (approved):
Nice 👍

PR #530, opened 2023-09-21
refactor: create TPackage and use in tree

ingeridhellen:

Creates the TPackage type and use in tree. Fix bugs that arise.

Typing helps us avoid bugs


henrikbossart (approved):
lgtm

PR #531, opened 2023-09-21
refactor(tree): remove onselect from wrapper props

ingeridhellen:

Moves the SelectPackageButton so that it doesn't have to get the onSelect prop through the TreeView. Then remove the now unused NodeWrapperOnClick prop from TreeView

The onSelect function was passed into the TreeView and directly out again to the caller. It was only used in the DestinationPicker


collinlokken (approved):
Looks good!

PR #532, opened 2023-09-21
test: improve view-selector test

roar-larsen:
Added "tab" as a role to the sidebar. When used this makes the playwright tests more readable.
Some other improvements to the test.

Chose to use "tab" as a role as that felt as the best fit.


henrikbossart (approved):
Looks fine, and test passes, sooooooo fine? 🤷‍♂️ 😄

PR #536, opened 2023-09-22
fix: temp fix for view-selector test

### LINKED ISSUE #527: View selector test fails wrongfully
`plugin-view_selector-car_garage.spec.ts` test fails, even though the test steps pass when doing it manually.

fails first at collapsing the sidebar.

roar-larsen:
Temporary fix that hopefully solves the view selector playwright test from failing.

It looks like a previous tooltip gets in the way when clicking the collapse button. Added an extra click to get the tooltip away.
image

Closes #527


collinlokken (approved):
Looking good!

Issue #534, opened 2023-09-21
fix(tree): rewrite tree and fix bugs


ingeridhellen:
#### Provide Examples

PR #537, opened 2023-09-22
refactor(tree): switch html structure from div to ul

ingeridhellen:

Before, the tree used divs upon divs. But actually, a nested list with buttons is a more correct approach, so I fixed that. I've also kept further improvements in mind while working with this, so I hope I easily can add new functionality on top of this.

Makes it easier to navigate in tests and increases accessibility. Hopefully, the code is easier to maintain too.

ref #534

### DIFF
--- a/packages/dm-core/src/components/TreeView.tsx
+++ b/packages/dm-core/src/components/TreeView.tsx
@@ -1,12 +1,11 @@
-import { useEffect, useState } from 'react'
+import { useState } from 'react'
 import styled from 'styled-components'
 import { TreeNode } from '../domain/Tree'
 
-import { Progress, Tooltip } from '@equinor/eds-core-react'
+import { Button, Icon, Progress, Tooltip } from '@equinor/eds-core-react'
+import { chevron_down, chevron_right } from '@equinor/eds-icons'
 import React from 'react'
 import {
-  FaChevronDown,
-  FaChevronRight,
   FaDatabase,
   FaExclamationTriangle,
   FaFileArchive,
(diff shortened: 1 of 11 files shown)


henrikbossart:
Screenshot 2023-09-22 at 15 05 33
This is totally a non-blocker, but could we drop the data source on these? it is already implicit as its located inside the data source directory (if that makes sense (or maybe I'm totally not getting how this works))

collinlokken (approved):
Good job!

packages/dm-core/src/components/TreeView.tsx
```diff
@@ -19,34 +18,28 @@ import {
...
 
-type TStyledTreeNode = {
-  level: number
-}
-
-const ExpandButton = styled.div`
-  margin: 0 3px;
-  width: 15px;
+const StyledButton = styled(Button)`
+  padding: 0 8px 0 0;
```
henrikbossart, line 22:
```suggestion
  padding-right: 8px;
```

henrikbossart: nvm,

henrikbossart (approved):
oh so pretty

ingeridhellen:
> Screenshot 2023-09-22 at 15 05 33 This is totally a non-blocker, but could we drop the data source on these? it is already implicit as its located inside the data source directory (if that makes sense (or maybe I'm totally not getting how this works))

I agree. However, that is not my doing. The TreeView just displays the name of the node, and if that does not exist, the id. Those values are set by the Tree, so that would be a different issue.

PR #538, opened 2023-09-22
fix(tree): add svg title

ingeridhellen:

Adds title to icons, thus removing the accessibility warning of missing text labels

Accessibility

ref #534


collinlokken (approved):
Nice!

PR #543, opened 2023-09-25
refactor: change the text on the job button to be start job

henrikbossart:
Changes the button label on the create job button to be "start job"
babak request


collinlokken (changes requested):
This is not really what we want. What Babak wants is to immediately be presented with the `JobControlButton` ("play"-button). It does not really make sense to have two different buttons for the same action. The start-handler in `JobControlButton`, should really do the same thing as the `Create job`-button (call the createNewJob-function on click).

henrikbossart:
@collinlokken I'vre removed the create job component, please have another look 👍

collinlokken (approved):
Perfect! Good job :)

PR #547, opened 2023-09-26
refactor(tree): remove unused level prop

ingeridhellen:

Removes level from TreeNode

Earlier, level was used to create correct indenting of the TreeView. This is no longer needed

ref #534


henrikbossart (approved):
lgtm 👍

PR #554, opened 2023-09-27
refactor(tree): remove attribute from treenode

ingeridhellen:

Removes `attribute` from TreeNode class.

Not really needed, and it increases the complexity of the code

#ref #534


eoaksnes (approved):
Nice

PR #561, opened 2023-09-29
fix: update view-selector test after EDS update

roar-larsen:
EDS added support for aria label to the collapse/expand button, which makes it a lot easier for Playwright to get the element.


ingeridhellen (approved):
🎉

PR #563, opened 2023-09-29
refactor(tree): change input to TreeNode

ingeridhellen:

By changing input to an object, the arguments can be non-positional

Makes the code easier to read

ref #534


henrikbossart (approved):
👌

Issue #495, opened 2023-09-11
Make the SIGNAL JOB UI plugin generalized so that it can be any job ui plugin.


pbullhove:

henrikbossart:
I guess the one we have today in the `example/src/pages` could be fairly general. Only thing to change required is to actually take in the addresses and such for the jobs.

PR #565, opened 2023-09-29
Feat: more generalized job plugin

henrikbossart:
Moves the job plugin from the example app into `dm-core-plugins` and extracts the config out of the plugin and into the recipe. I'm not 100% sure my thinking here about the recipes and blueprints are correct, so please arrest me.

I moved the old job stuff into a new dir `job_old` for the time being, as I'm not sure if there is something we can reuse in there. Should be removed later though.

Same goes for the job plugin in the example app. Its still there, but only for reference. It should be removed before this PR is merged

requested feature

#495


packages/dm-core-plugins/src/index.tsx
```diff
@@ -41,12 +43,15 @@ export default {
...
-    component: JobInputEditPlugin,
+  '@development-framework/dm-core-plugins/job/single_job': {
+    component: JobPlugin,
   },
+  // '@development-framework/dm-core-plugins/jobControl': {
+  //   component: JobControlPlugin,
+  // },
+  // '@development-framework/dm-core-plugins/jobInputEdit': {
+  //   component: JobInputEditPlugin,
+  // },
```
collinlokken, line 54: same here

packages/dm-core-plugins/src/index.tsx
```diff
@@ -3,8 +3,10 @@ import { TUiPluginMap } from '@development-framework/dm-core'
 import { YamlPlugin } from './yaml/YamlPlugin'
 import { ViewSelectorPlugin } from './view_selector/ViewSelectorPlugin'
 import { BlueprintHierarchyPlugin } from './blueprint-hierarchy/BlueprintHierarchyPlugin'
-import { JobInputEditPlugin } from './job/JobInputEditPlugin'
-import { JobControlPlugin } from './job/JobControlPlugin'
+// Deprecated?
+// import { JobInputEditPlugin } from './job_old/JobInputEditPlugin'
+// import { JobControlPlugin } from './job_old/JobControlPlugin'
```
collinlokken, line 8: probably just remove the imports, it is easy to import them back if we need

packages/dm-core-plugins/package.json
```diff
@@ -24,16 +25,16 @@
...
     "jest-environment-jsdom": "^29.5.0",
     "react-router-dom": ">=5.1.2",
     "shx": " 0.3.4",
     "ts-jest": "^29.1.0",
     "ts-loader": "^9.3.1",
-    "typescript": "^5.1.6",
-    "eslint-plugin-react": "^7.33.2",
-    "@typescript-eslint/eslint-plugin": "^6.7.3",
-    "@typescript-eslint/parser": "^6.7.3"
+    "typescript": "^5.1.6"
```
collinlokken, line 37: switcheroo? personal preference, but I would probably not include this in the PR since it is not related.

example/app/data/DemoDataSource/recipes/apps/signal_app/recipe_links/signals_simple/case.recipe.json
```diff
@@ -98,10 +98,30 @@
       }
     },
     {
-      "name": "RunInt",
+      "name": "RunJob",
       "type": "CORE:UiRecipe",
       "description": "running",
-      "plugin": "signal-job-single"
```
collinlokken, line 104: maybe remove this from `example/src/plugins/job-ui-single/index.tsx` and from `example/src/plugins/index.tsx`?

packages/dm-core/package.json
```diff
@@ -14,25 +14,26 @@
     "@equinor/eds-icons": "^0.19.3",
     "@equinor/eds-tokens": "^0.9.0",
     "axios": "^1.4.0",
+    "lodash": "^4.17.21",
     "react-icons": "4.10.1",
-    "react-toastify": "^9.1.3",
-    "react-oauth2-code-pkce": "^1.10.1"
+    "react-oauth2-code-pkce": "^1.10.1",
+    "react-toastify": "^9.1.3"
```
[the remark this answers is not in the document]
henrikbossart: Nothing I've done myself, so I'm guessing its the IDE that did it at some point 🤷‍♂️ Doesn't really matter, but I can change it back if its a problem

PR #571, opened 2023-10-02
ci: deploy to Radix

eoaksnes:

* Fixed so that the packages are not tried to be published twice and therefore fail the second time (since it is already uploaded)
* Added the possibility to manually trigger the `on-release` workflow
* Added so that release please is responsible for triggering the  `on-release` after a release is created


collinlokken (approved):
Looking good!

PR #575, opened 2023-10-03
feat: Example application with role selector and role filter

### LINKED ISSUE #568: feat(example): use the header plugin and have proper application entity
The application is now hard coded to start the `App.tsx` that contains a Tree component and EntityView. The content of the `App.tsx` should be ported to be a plugin, and instead, an application blueprint should point to a header plugin that sets the content to point to this new plugin.

```
src/plugins/<some-new-plugin>
```

The goal is to have the example app identical to how the create-dm-app works, some of the changes needed are:

* The `App.tsx` should be like this https://github.com/equinor/create-dm-app/blob/main/src/App.tsx
* The `.env` file points to the application entity.
* Need to create a blueprint for Example app like this: https://github.com/equinor/create-dm-app/blob/main/app/data/DemoApplicationDataSource/DemoApplication/blueprints/DemoApplication.json
* Need to create a entity for the example app like this: https://github.com/equinor/create-dm-app/blob/main/app/data/DemoApplicationDataSource/DemoApplication/entities/demoApplication.json
* +++

collinlokken:
Set application wide role by:
- Moves role selection into HeaderPlugin which sets an application wide role, which can be used in the role filter.
- Generalize App.tsx to initialise application with entity specified in env. instead of hardcoded Tree + EntityView
- Initialise application using Example Application entity that uses HeaderPlugin as default, which points to the ExplorerPlugin

Allows roles from Azure to be used in the app to deliver specific views for specific users/roles

Closes #568


packages/dm-core-plugins/package.json
henrikbossart, line 1: Funny, you commented on this exact change in my PR 😅

packages/dm-core-plugins/package.json
```diff
@@ -10,7 +10,6 @@
     "@equinor/eds-tokens": "^0.9.0",
     "axios": "^1.4.0",
     "highlight.js": "^11.8.0",
-    "lodash": "^4.17.21",
```
henrikbossart, line 13: Are you removing it again? 🤔

henrikbossart (commented):
I only get one role in the modal, how do I get more? looks like the roles come from the `tokenData`?

packages/dm-core-plugins/package.json
collinlokken: It was just supposed to remove lodash!! 😅

packages/dm-core-plugins/package.json
collinlokken: Yes, it is no longer required :)

soofstad (dismissed):

packages/dm-core-plugins/src/role_filter/RoleFilterPlugin.tsx
```diff
@@ -87,7 +75,7 @@ export const RoleFilterPlugin = (props: IUIPlugin): JSX.Element => {
             <Icon name="thumbs_down" />
           </Banner.Icon>
           <Banner.Message>
-            {`No views found, since you currently have roles [${User.roles}]. Please switch to one of these roles: [${allowedRoles}]`}
+            {`No views found, since you currently have role [${selectedRole}]. Please switch to one of these roles: [${allowedRoles}]`}
```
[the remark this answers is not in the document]
collinlokken: I understand what you mean. Right now you can define a view without any roles, which will show for users without a role (or with a wrong role). Like this you are not forced to show anything, which is nice imo. If we fallback to e.g. the yaml plugin, now we have introduced a dependency, which sounds like something we want to avoid. I am still not super confident on the plans for the future, but it would be a problem if say you dont want to install the yaml plugin (if that is a use case). Any thoughts on this @eoaksnes @henrikbossart @pbullhove ?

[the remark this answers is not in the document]
henrikbossart: Maybe the fallback can be an optional parameter? I can imagine cases where an application would use a fallback in case of lacking access permissions, but I can also see cases where an application might choose to just show an "you don't have access to this resource/view/entity" disclaimer. 

I do agree about it being hard to say anything about the future needs of the framework though 😅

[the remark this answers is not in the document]
soofstad: Agreed we keep the flexibility of optional fallback in the config for RolerFilterPlugin :+1:

eoaksnes (approved):

PR #576, opened 2023-10-03
feat: add sample json schema for validating blueprints and recipes

henrikbossart:
Here's a "quick" demo of what I talked about last week regarding json-schema and validation. Its probably a bit of a set-up job, but I think this is something that could prove pretty useful for us, and for other people using the framework.

This is purely a DX feature, and has to be manually enabled in the editor (afaik).

To enable the json schema validation:

For pycharm/intellij/webstorm, open settings, and go to `Languages & Frameworks` -> `Schemas and DTDs` -> `JSON Schema Mappings` and add a schema for the recipe and one for the blueprint schema. Then, make it validate files matching the paths `*.recipe.json` and `*.blueprint.json`

in VSCode, open settings and search for JSON Schema. It will tell you to edit `settings.json`. There, you'll have to add the following snippet to link to the schemas:
```json
{
    "json.schemas": [
        {
            "fileMatch": [
                "*.blueprint.json"
            ],
            "url": "file:<PATH TO dm-core-packages>/_schema/blueprint.schema.json"
        },
        {
            "fileMatch": [
                "*.recipe.json"
            ],
            "url": "file:<PATH TO dm-core-packages>/_schema/recipe.schema.json"
        }
    ]
}
```

If set up correctly you'll now get suggestions when writing blueprints and recipes, and errors if you're missing something or have provided an invalid value.
Screenshot 2023-10-03 at 13 45 05
Screenshot 2023-10-03 at 13 45 17

### DIFF
--- a/_schema/blueprint.schema.json
+++ b/_schema/blueprint.schema.json
@@ -1,0 +1,74 @@
+{
+  "type": "object",
+  "description": "Schema for writing correct blueprints",
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "Blueprint schema",
+  "additionalItems": false,
+  "$defs": {
+    "type": {
+      "type": "string",
+      "enum": [
+        "CORE:Blueprint",
+        "dmss://system/SIMOS/Blueprint"
+      ]
+    },
+    "attributeTypes": {
+      "type": "string",
+      "enum": [
+        "CORE:BlueprintAttribute",
+        "dmss://system/SIMOS/BlueprintAttribute"
+      ]
+    }
+  },
+  "properties": {
+    "name": {
+      "type": "string"
+    },
+    "type": {
+      "$ref": "#/$defs/type"
+    },
+    "description": {
+      "type": "string"
+    },
+    "attributes": {
+      "type": "array",
+      "items": {
+        "type": "object",
+        "properties": {
+          "name": {
+            "type": "string"
+          },
+          "type": {
+            "$ref": "#/$defs/attributeTypes"
+          },
+          "description": {
+            "type": "string"
+          },
+          "attributeType": {
+            "type": "string"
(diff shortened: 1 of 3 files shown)


_schema/blueprint.schema.json
```diff
@@ -0,0 +1,74 @@
...
+            "type": "boolean"
+          },
+          "label": {
+            "type": "string"
+          },
+          "dimensions": {
+            "type": "string"
+          },
+          "default": {
+            "type": "string"
```
soofstad, line 60: The type here is "any"

soofstad (approved):
Cool. Definitely worth testing out.

One small comment

_schema/blueprint.schema.json
henrikbossart: makes sense

henrikbossart:
Right, this is by no means a finished and complete schema. I'm far from familiar enough with all the possibilities/types/configs/... with the blueprints and stuff, so this was merely meant as a demo for others. Ideally, we would update the schemas when creating new blueprints and recipes.

What would also be cool is if we could add schemas for all the various plugins, so that the end users get type hinting in when configuring everything. Not sure how feasible that is though..

_schema/blueprint.schema.json
henrikbossart, line 60:
```suggestion
```

_schema/blueprint.schema.json
```diff
@@ -0,0 +1,73 @@
...
+            "type": "boolean"
+          },
+          "label": {
+            "type": "string"
+          },
+          "dimensions": {
+            "type": "string"
+          },
+          "default": {
+          }
```
henrikbossart, line 60:
```suggestion
          "default": true
```

PR #579, opened 2023-10-05
Feat/application header test mode

collinlokken:
Introduces manually setting roles in testing environment

Roles can be manually set without actually adding them in Azure


soofstad (approved):
Looks good! :+1:

Issue #572, opened 2023-10-02
feat: allow to specify the dm job url inside the .env file instead of hardcoded in the package.json script


eoaksnes:
The URL should be specified similar to the DMSS URL, inside the .env file. 

Then, the job URL needs to be provided similar to the DMSS, something like this:

```
<DMSSProvider dmssBasePath={import.meta.env.VITE_DMSS_URL}>
````

See `src/index.tsx` file for more code.... 

We might need a JobProvider or combine these ting into a common provider (or the application provider).

PR #580, opened 2023-10-05
feat: .env variable and context for dm-job api

henrikbossart:
Adds a context provider for the dm-job api, and takes in the url for it through the env

Note: I havn't populated other env files than the local one, as I don't know what the URL for the job api is.

#572


packages/dm-core/src/hooks/useJob.tsx
```diff
@@ -82,8 +81,9 @@ export function useJob(entityId?: string, jobId?: string): IUseJob {
     typeof setInterval
   > | null>(null)
   const [error, setError] = useState<ErrorResponse>()
-  const { token } = useContext(AuthContext)
-  const dmJobApi = new DmJobAPI(token)
+  // const { token } = useContext(AuthContext)
+  // const dmJobApi = new DmJobAPI(token)
```
soofstad, line 85: Can probably remove this

soofstad (approved):

packages/dm-core/src/hooks/useJob.tsx
henrikbossart: oops

henrikbossart, line 85:
```suggestion
```

PR #587, opened 2023-10-06
refactor: change return type of useDocument

henrikbossart:
Changes the return type of the `useDocument` hook to avoid unused variables.
cleanup


soofstad (approved):
Nice. This one had a long time coming :smile:

PR #594, opened 2023-10-10
feat: storage uncontained demo

eoaksnes:

* Update `dm-cli` version that can handle global files (needed for storage uncontained)
* Updated the already existing storage uncontained example in form


collinlokken (approved):
Looks good 👍

Issue #613, opened 2023-10-11
bug: viewing a list entity does not select the correct uiRecipe for dimensions "*"


soofstad:
#### Expected Behaviour
Clicking on a list in the Explorer should show a list of the items

#### Actual Behaviour

We get to see the list entity with the uiRecipes meant to render a single entity

#### Steps to Reproduce

henrikbossart:
Isn't this happening because of how the recipes are configured? 🤔

PR #626, opened 2023-10-13
fix: render lists as lists

henrikbossart:
Makes sure that list entities are rendered as lists.
Not sure this approach is sufficient, but its something..

To render lists as lists

#613


soofstad (approved):
Perfect! :smile:

PR #635, opened 2023-10-17
feat: minor adjustment to tree for how to show references and attributes

eoaksnes:

* Only show the expandable icon if the node has any children that are complex (or array)
* Show special icon for references
* Show attribute names instead of name of entity for complex attributes
* Add example for list of model uncontained to check that list of references also get correct icon

From

image

To

image


henrikbossart (approved):
Awesome! 🙌

PR #637, opened 2023-10-17
refactor: properly support datepicker in form

henrikbossart:
Adds proper support for dates in the form.
The date string is formatted after the ISO 8601 spec.

To make the form render a datepicker, use the following config (replace `"name": "date"` with whats appropriate for your use-case):

```Json
...
{
  "config": {
    ...
    "attributes": [
      {
        "name": "date",
        "type": "PLUGINS:dm-core-plugins/form/fields/StringField",
        "format": "datetime-local"
      }
    ]
  }
}
...
```
The format has to be one of the formats specified by EDS: `"date" | "time" | "datetime-local"`. Note that if `"time"` is used, the value cannot be provided on the ISO format, instead the time should be provided on the 24h format.

We probably want a standalone component for the date picker as well, which can be used outside the form(?).


example/app/data/DemoDataSource/plugins/form/simple/simple.entity.json
```diff
@@ -4,5 +4,6 @@
   "stringRequired": "",
   "numberRequired": 2,
   "stringOptional": "a prefilled optional string",
-  "requiredCheckbox": false
+  "requiredCheckbox": false,
+  "date": "2023-10-17T13:30"
```
soofstad, line 8: Should TZ be part of the timestamp?

packages/dm-core-plugins/src/form/fields/AttributeField.tsx
```diff
@@ -114,6 +115,17 @@ export const AttributeField = (props: TAttributeFieldProps) => {
...
+    //   return (
+    //     <DatePicker
+    //       namePath={namePath}
+    //       displayLabel={displayLabel}
+    //       defaultValue={attribute.default}
+    //       optional={attribute.optional ?? false}
+    //       uiAttribute={uiAttribute}
+    //       readOnly={readOnly}
+    //     />
+    //   )
```
soofstad, line 128:
```suggestion
```

soofstad (approved):

Issue #641, opened 2023-10-18
fix: JobPlugin does not stop rerendering and calling the API


collinlokken:
A bug in the JobPlugin makes it not stop rerendering, which in turn makes it call the API

![Image](https://github.com/equinor/dm-core-packages/assets/43639886/4151f810-50c4-4316-bde5-2eb06d3f8af4)

![Image](https://github.com/equinor/dm-core-packages/assets/43639886/c727bf37-107b-47bb-8db6-29e9ec6bd742)

Steps to reproduce:
1. Run the dm-simpos-app
2. Go to MSDS->ASGB->ASGB->Edit->omnia_ts_jobs-> expand first list item->Run the job

Expected behaviour:
The JobPlugin returns to ready-state and waits for further instructions after the job has finished

Actual behaviour:
The JobPlugin keeps sending GET requests to the job-api after the job has finished

PR #642, opened 2023-10-18
fix: job pinger never stops

henrikbossart:
Changes how the intervalID of the job pinger is stored. It now uses a ref instead of state, which doesn't trigger a re-render.

This change makes the pinger stop when the state is not `"running"`

#641


collinlokken (approved):
Nice! Looks like useRef is very appropriate for this case 👍

soofstad (approved):

PR #645, opened 2023-10-19
fix: make grid columns span correctly

### LINKED ISSUE #638: fix(grid): columnEnd is off by one
#### Expected Behaviour

Given a grid 2x2; I expect to define a view to end in the last column to have `"columnEnd": 2`.

#### Actual Behaviour
"columnEnd" needs to be set to "3" for the view to take the whole width.

henrikbossart:
the grid positioning is referencing the grid lines, not the columns. Therefor I increment the end by one to reference the "ending line" of the column
closes #638

### DIFF
--- a/packages/dm-core-plugins/src/grid/GridPlugin.tsx
+++ b/packages/dm-core-plugins/src/grid/GridPlugin.tsx
@@ -21,9 +21,6 @@ export const GridPlugin = (props: TGridPluginConfig): React.ReactElement => {
       rowGap={config.size.rowGap}
       columnGap={config.size.columnGap}
     >
-      {config.items.forEach(() => (
-        <div style={{ backgroundColor: 'hotpink' }}></div>
-      ))}
       <GridItems idReference={idReference} items={config.items} type={type} />
     </Grid>
   )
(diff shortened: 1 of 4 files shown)


henrikbossart:
That didn't quite turn out how I wanted...

henrikbossart:
Slightly tweaked an example, still not 100%, but I think the grid works as intended now, so the examples shouldn't be blocking the fix for other use-cases.

packages/dm-core-plugins/src/grid/GridPlugin.tsx
```diff
@@ -21,6 +21,9 @@ export const GridPlugin = (props: TGridPluginConfig): React.ReactElement => {
       rowGap={config.size.rowGap}
       columnGap={config.size.columnGap}
     >
+      {config.items.forEach(() => (
+        <div style={{ backgroundColor: 'hotpink' }}></div>
+      ))}
```
soofstad, line 26: Whats this? :thinking:

soofstad (approved):

packages/dm-core-plugins/src/grid/GridPlugin.tsx
henrikbossart, line 26:
```suggestion
```

PR #646, opened 2023-10-19
feat: support for scheduling of cron jobs in JobPlugin

collinlokken:
Adds support for cron job in JobPlugin

Makes it possible to schedule the job for later

### DIFF
--- a/packages/dm-core-plugins/src/job/CronJob.tsx
+++ b/packages/dm-core-plugins/src/job/CronJob.tsx
@@ -3,7 +3,7 @@ import { Button, Label } from '@equinor/eds-core-react'
 import { StyledSelect } from './Input'
 import DateRangePicker from './DateRangePicker'
 import styled from 'styled-components'
-import { TGenericObject, TSchedule } from '@development-framework/dm-core'
+import { TSchedule } from '@development-framework/dm-core'
 
 enum EInterval {
   HOURLY = 'Hourly',
@@ -40,14 +40,14 @@ export function CreateReoccurringJob(props: {
   removeJob: () => void
   // TODO: Export TCronJob from dm-core
   setCronJob: (job: TSchedule) => void
-  cronJob?: TGenericObject | undefined
+  cronJob?: TSchedule | undefined
 }) {
   const { close, removeJob, setCronJob, cronJob } = props
   const [schedule, setSchedule] = useState<string | null>(null)
   const [dateRange, setDateRange] = useState<{
     startDate: string
     endDate: string
-  }>({ startDate: '', endDate: '' })
+  }>({ startDate: cronJob?.startDate ?? '', endDate: cronJob?.endDate ?? '' })
   const [interval, setInterval] = useState<EInterval>(EInterval.HOURLY)
   const [hour, setHour] = useState<string>('23')
   const [hourStep, setHourStep] = useState<string>('1')
@@ -176,7 +176,6 @@ export function CreateReoccurringJob(props: {
               //@ts-ignore
               endDate: dateRange.endDate,
             })
-            close()
           }}
         >
           Set
--- a/packages/dm-core-plugins/src/job/CronJob.tsx
+++ b/packages/dm-core-plugins/src/job/CronJob.tsx
@@ -1,0 +1,187 @@
+import React, { useEffect, useState } from 'react'
+import { Button, Label } from '@equinor/eds-core-react'
+import { StyledSelect } from './Input'
+import DateRangePicker from './DateRangePicker'
+import styled from 'styled-components'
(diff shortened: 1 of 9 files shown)


packages/dm-core/src/types.ts
```diff
@@ -100,6 +100,14 @@ export type TJob = {
   ended?: string
   outputTarget?: string
   referenceTarget?: string
+  schedule?: TSchedule
+}
+
+export type TSchedule = {
+  type: string
```
henrikbossart, line 107: Are we planning on having different kinds of schedules? 🤔

henrikbossart (approved):

soofstad (changes requested):
Instead of going multiple rounds on this.
I suggest we bring in the old "cronJob"-code and do it properly the first time.
https://github.com/equinor/dm-app-analysis-platform/blob/main/src/plugins/analysis-platform/modules/Jobs/not_used/ReoccurringJob.tsx
There is no big rush getting this into the frontend yet

packages/dm-core-plugins/src/job/DateRangePicker.tsx
```diff
@@ -0,0 +1,40 @@
+import React from 'react'
+import { TextField } from '@equinor/eds-core-react'
+
+const DateRangePicker = (props: {
+  setDateRange: (dateRange: { startDate: string; endDate: string }) => void
+  value: any
+}): JSX.Element => {
+  const { setDateRange, value } = props
+
+  console.log(value)
```
eoaksnes, line 10: Please remove console.log

packages/dm-core-plugins/src/job/JobPlugin.tsx
```diff
@@ -117,12 +141,37 @@ export const JobPlugin = (props: IUIPlugin & { config: JobPluginConfig }) => {
...
+        }
+        checked={asCronJob}
+      />
+      {asCronJob && (
+        <CreateReoccurringJob
+          cronJob={jobSchedule}
+          close={() => setAsCronJob(false)}
+          removeJob={() => setJobSchedule(emptyJob)}
+          setCronJob={(e: TSchedule) => {
+            console.log(e)
```
eoaksnes, line 163: Please remove console.log

packages/dm-core-plugins/src/job/CronJob.tsx
```diff
@@ -0,0 +1,186 @@
...
+          setDateRange={(dateRange) => setDateRange(dateRange)}
+          value={dateRange}
+        />
+        <InputWrapper>
+          <div style={{ display: 'flex', flexDirection: 'column' }}>
+            <Label label="Interval" />
+            <StyledSelect
+              onChange={(e) => {
+                //@ts-ignore
+                // TODO fix type
```
eoaksnes, line 104: Suggestion: Would be nice if we can remove the ts-ignore.

packages/dm-core-plugins/src/job/CronJob.tsx
```diff
@@ -0,0 +1,186 @@
...
+          )}
+          {interval === EInterval.HOURLY && (
+            <div style={{ display: 'flex', flexDirection: 'column' }}>
+              <Label label="Hour step" />
+              <StyledSelect
+                onChange={(e) => {
+                  setHourStep(e.target.value)
+                }}
+              >
+                {/*TODO fix type error*/}
```
eoaksnes, line 143: Suggestion: Would be nice if we can remove the ts-ignore.

eoaksnes (approved):
Two console.log. and suggest to remove the ts-ignore

packages/dm-core-plugins/src/job/JobPlugin.tsx
```diff
@@ -79,12 +83,7 @@ export const JobPlugin = (props: IUIPlugin & { config: JobPluginConfig }) => {
...
   const [asCronJob, setAsCronJob] = useState<boolean>(false)
-  const [jobSchedule, setJobSchedule] = useState<TSchedule>({
-    type: EBlueprint.CRON_JOB,
-    cron: defaultCron,
-    startDate: '',
-    endDate: '',
-  })
+  const [jobSchedule, setJobSchedule] = useState<TSchedule>(emptyJob)
   const canSubmit =
     !asCronJob || Boolean(jobSchedule.startDate && jobSchedule.endDate)
```
soofstad, line 88: Know this is outside current PR.
But, no need to set this requirement on start- endDate when we don't use them for anything

soofstad (approved):

packages/dm-core-plugins/src/job/JobPlugin.tsx
collinlokken: We can disable the check, but I suggest that we do it in another PR

PR #649, opened 2023-10-23
fix(jobPlugin): multiple issues with jobPlugin

### LINKED ISSUE #647: bug(JobPlugin): Wrong address inserted into jobAddress in JobPlugin
Wrong address inserted into jobAddress in JobPlugin

soofstad:
- Some refactoring and code cleanup
- JobControlButton now correctly show proper action that the user can take
- JobControlButton no longer remove jobs
-  'useDocument's update function is now async, with a "notify" parameter
- "useJob" now ties the status interval with the "jobHookId"

- Fixes multiple bugs with "useJob" and JobPugin

closes #647

### DIFF
--- a/packages/dm-core-plugins/src/job/JobControlButton.tsx
+++ b/packages/dm-core-plugins/src/job/JobControlButton.tsx
@@ -5,20 +5,13 @@ import React, { MutableRefObject, useRef, useState } from 'react'
 
 export const JobControlButton = (props: {
   jobStatus: JobStatus
-  jobExists: boolean
-  createJob: () => Promise<unknown>
-  start: () => void
+  createJob: () => void
 }) => {
-  const { jobStatus, start, jobExists, createJob } = props
+  const { jobStatus, createJob } = props
   const [hovering, setHovering] = useState(false)
   const buttonRef: MutableRefObject<HTMLButtonElement | undefined> = useRef()
   buttonRef.current?.addEventListener('mouseenter', () => setHovering(true))
   buttonRef.current?.addEventListener('mouseleave', () => setHovering(false))
-
-  function handleClick() {
-    if (!jobExists) return createJob().then(() => start())
-    start()
-  }
 
   const buttonText = () => {
     switch (jobStatus) {
@@ -47,7 +40,7 @@ export const JobControlButton = (props: {
   }
 
   return (
-    <Button ref={buttonRef} onClick={handleClick} style={{ width: '7rem' }}>
+    <Button ref={buttonRef} onClick={createJob} style={{ width: '7rem' }}>
       {jobStatus === JobStatus.Running && !hovering ? (
         <CircularProgress size={16} variant="indeterminate" />
       ) : (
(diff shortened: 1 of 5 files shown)


henrikbossart:
This is not working for me. It fails to get the job attribute.
```
{
  "status": 404,
  "type": "NotFoundException",
  "message": "Failed to get document referenced with 'dmss://DemoDataSource/$4483c9b0-d505-46c9-a157-94c79f4d7a6a.study.cases[0].job'",
  "debug": "Invalid attribute 'job'. Valid attributes are '['name', 'type', 'description', 'duration', 'timeStep', 'components', 'signal']'.. The requested resource could not be found",
  "data": null
}


```

packages/dm-core-plugins/src/job/JobControlButton.tsx
```diff
@@ -1,53 +1,57 @@
...
-      if (!jobExists) {
-        createJob().then(() => {
-          start()
-        })
-      } else {
-        start()
-      }
-    }
+  function handleClick() {
+    if (!jobExists) return createJob().then(() => start())
```
collinlokken, line 19: `createJob` allready handles `jobExists`, and either updates or creates new. Right now it is not possible to update the job if it already exist. handleClick should just call `createJob` then `start`

packages/dm-core-plugins/src/job/JobPlugin.tsx
```diff
@@ -90,86 +87,34 @@ export const JobPlugin = (props: IUIPlugin & { config: JobPluginConfig }) => {
...
-        setJobExists(true)
-      })
-    }
-  }, [jobEntityId])
-
   function createNewJob(): Promise<unknown> {
-    setAllowJobStart(true)
     if (jobExists) {
-      return updateDocument(jobTargetAddress(), jobEntity)
+      return updateDocument(jobEntity, false).then(() => setJobExists(true))
```
collinlokken, line 108: Do we need to setJobExistst -> true here?  It is already inside the `if...` because `jobExists`

packages/dm-core-plugins/src/job/JobPlugin.tsx
```diff
@@ -44,43 +45,39 @@ export const JobPlugin = (props: IUIPlugin & { config: JobPluginConfig }) => {
...
-    start,
-    error,
-    fetchResult,
-    fetchStatusAndLogs,
-    logs,
-    status,
-    remove,
-  } = useJob(jobEntityId, jobId)
+    document: jobDocument,
+    isLoading,
```
collinlokken, line 74: `isLoading: bool`

collinlokken (changes requested):
Nice! With some comments :)

packages/dm-core-plugins/src/job/JobPlugin.tsx
henrikbossart: not really needed? and the type would be `Boolean` ;)

soofstad:
@henrikbossart Yes that was showing for 1sec before it continued.
Fixed it by resetting error in useJob :+1:

henrikbossart (approved):
:+1:

collinlokken (approved):
Nice

PR #651, opened 2023-10-23
refactor: add time zone and remove unused code

henrikbossart:


soofstad (approved):
:+1:  Good!

henrikbossart:
@soofstad I had to add a dependency to be able to convert the ISO to a local date object without the time zone. I know you have feelings about bundle-size, so please have a look :)

soofstad (approved):

PR #657, opened 2023-10-24
Fix/tree correct node id prefix

soofstad:
- Fixes a bug where nodeId in the Tree did not have the "dmss://" prefix
- Feature: expose the EntityPickerDialog "scope" parameter for ListPlugin

- Frontend generated invalid references because of wrong nodeID
- When selecting items to append to a list, should be possible to limit the tree view

none


collinlokken (approved):
Nice!

henrikbossart (approved):

PR #664, opened 2023-10-26
refactor(list): use list hook

eoaksnes:

* Use the `useList` hook inside `ListPlugin` to handle list operations

### DIFF
--- a/packages/dm-core/src/hooks/useList.tsx
+++ b/packages/dm-core/src/hooks/useList.tsx
@@ -1,65 +1,81 @@
 import { ErrorResponse } from '../services'
 import { useEffect, useState } from 'react'
 import { useDMSS } from '../context/DMSSContext'
-import { AxiosError, AxiosResponse } from 'axios'
-import {
-  TAttribute,
-  TGenericObject,
-  TLinkReference,
-  TValidEntity,
-} from '../types'
-import { toast } from 'react-toastify'
+import { AxiosError, AxiosResponse, isAxiosError } from 'axios'
+import { TAttribute, TLinkReference } from '../types'
 import { EBlueprint } from '../Enums'
 
 export type TItem<T> = {
   key: string
   index: number
-  data: T
+  data: T | null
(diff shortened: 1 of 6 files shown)


packages/dm-core/src/hooks/useList.tsx
```diff
@@ -52,115 +67,152 @@ export function useList<T>(idReference: string): IUseListReturnType<T> {
...
+          if (attribute.contained) {
+            if (!Array.isArray(response.data)) {
+              throw new Error(
+                `Not an array! Got document ${JSON.stringify(response.data)} `
+              )
+            }
+
             const items = Object.values(response.data).map((data, index) => ({
-              key: crypto.randomUUID(),
+              key: uuidv4(),
```
soofstad, line 96: Why use extra package for uuid when it's supported in stdlib?

soofstad:
Would like to test before merging

packages/dm-core/src/hooks/useList.tsx
eoaksnes: Done 👍 Let's just use the default crypto instead.

henrikbossart:
Not working for me, I get no data when trying to load the examples.

i'm also getting the most descriptive error message in the world:
Screenshot 2023-10-31 at 10 26 44

soofstad:
Yepp, just tested the same:
Screenshot_20231031_103247

soofstad (approved):
Tested okey.
Seems to work perfectly :+1:

henrikbossart (changes requested):
This might not be related to this PR, but flagging it as 'request changes' to get attention on it:

If you move the last item one up, then save, then delete the last item, the reference is goneit seems.. 
I'll upload a screencast

henrikbossart:
https://github.com/equinor/dm-core-packages/assets/3105885/32d37473-e95d-4344-9d2c-d0a5f60afffc

henrikbossart (approved):
👍

PR #665, opened 2023-10-27
feat(ViewCreator): resolve references in view config

soofstad:
- Update generated TS DMSS API
- Change ViewCreator to resolve targets that are references by default

Most of the time, when rendering a child, we want to view the actual data, not the reference

none


henrikbossart (approved):
Seems fine :+1:

eoaksnes (approved):

PR #667, opened 2023-10-30
Feat/resizable explorer

### LINKED ISSUE #569: feat(explorer): make it possible to see the whole tree when expanding it
Image

How to solve?

- [ ] Should be possible to move the vertical line?
- [ ] Should be possible to have scrollbar on hortizontal for the tree?

henrikbossart:
Makes the explorer resizable. Also persists the width in the localstorage.
the explorer sometimes hides the elements.

closes #569 

https://github.com/equinor/dm-core-packages/assets/3105885/a19f6b55-b49b-42b1-8174-536da8741eb2


packages/dm-core-plugins/src/explorer/components/Sidebar.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+const [minWidth, maxWidth, defaultWidth] = [200, 500, 250]
+
+type TProps = {
+  children: ReactNode
+}
+
+export default function Sidebar(props: TProps) {
+  const { children } = props
+  const [width, setWidth] = useState<number>(
```
soofstad, line 11: We have a hook for this: useLocalStorage

packages/dm-core-plugins/src/explorer/components/Sidebar.tsx
```diff
@@ -0,0 +1,57 @@
...
+        style={{
+          width: '.1rem',
+          backgroundColor: 'gray',
+          cursor: 'col-resize',
+        }}
+        onMouseDown={() => {
+          isResized.current = true
+        }}
+      />
+    </div>
```
soofstad, line 55: Can we make this a bit bigger?
A bit hard to "catch". Does not need to be visually wider

packages/dm-core-plugins/src/explorer/ExplorerPlugin.tsx
```diff
@@ -4,20 +4,23 @@ import {
...
   TreeView,
 } from '@development-framework/dm-core'
-import { Progress } from '@equinor/eds-core-react'
 import React, { useContext, useState } from 'react'
 import styled from 'styled-components'
+import Sidebar from './components/Sidebar'
 import NodeRightClickMenu from './components/context-menu/NodeRightClickMenu'
+import { Progress } from '@equinor/eds-core-react'
 
 export const TreeWrapper = styled.div`
```
soofstad, line 13: This is no longer in use? 
If so should be removed

soofstad (changes requested):

packages/dm-core-plugins/src/explorer/components/Sidebar.tsx
```diff
@@ -8,8 +9,9 @@ type TProps = {
 
 export default function Sidebar(props: TProps) {
   const { children } = props
+  const sideBarWidth = useLocalStorage('explorerWidth', defaultWidth)
```
soofstad, line 12: This does not seem to be used correctly. 
You can use it like a normal "useState"-hook. It returns  [ value, setValue ].

Meaning you can replace your useState hook with this.

soofstad (changes requested):

packages/dm-core-plugins/src/explorer/components/Sidebar.tsx
henrikbossart: Can't get it to work... The value seems to be always writing the same.

soofstad (approved):

PR #668, opened 2023-10-30
ci: increase timeout for e2e tests

eoaksnes:

* Increase timeout for tests to be 60 seconds (default is 30 seconds)


henrikbossart (approved):

roar-larsen:
@eoaksnes I added `slowMo: 500` to the config, which seems to work. I have tested it quite a bit locally and it runs green every time now. However, this does slow down the test execution, especially in the CI where we only have 1 worker.

PR #675, opened 2023-10-31
build: cache node_modules

soofstad:
- Adds caching to some of the steps in the CI

- Increase CI speed, reduce cost, etc.

Linting and testing time is now 50% faster.
Which does not matter much since we are still waiting 7min on integration tests...


henrikbossart (approved):
lovely.

PR #676, opened 2023-10-31
refactor(e2e): utilize test.step in more tests

roar-larsen:
Utilising test.steps in more e2e-tests instead of using comments.


henrikbossart (approved):
That seems more appropriate, yes 👍

PR #677, opened 2023-10-31
refactor: fix typo

henrikbossart:
Fix typo: "Reoccurring" -> "Recurring"


soofstad (approved):
Both ways of writing it is allowed as far as I can gather.

So, whichever is fine :man_shrugging:

henrikbossart:
> Something that is recurring happens over and over again, possibly at regular intervals. In contrast, something that is reoccurring is simply happening again but not always repeatedly.

PR #681, opened 2023-11-01
fix(blueprint): fix several minor bugs in blueprint plugin

soofstad:
- Set "blueprint" as the default plugin for "Blueprint"-entities
- Reset value of "default" when changing type of the attribute
- Use a switch to set default value of a boolean
- Only render "contained"-switch on none-primitive types (excluding "any")

- Fixes several bugs with the plugin

none


eoaksnes (approved):
👍 smooth

PR #682, opened 2023-11-01
test: make e2e stable and faster

eoaksnes:

* Use production build (dist/) when running tests
* Cache playwright binary dependencies 
* Remove `slowMo` from playwright config
* Added extra expect that seems to make the code more stable (not 100% sure here)
* For `TabPlugin` and `SidebarPlugin`, don't render hidden tabs or sidebar content if not visible.
* Don't set loading false in `useList` plugin after getting the attribute, this makes it only disable loading after both attribute and the actual data is fetched

### DIFF
--- a/packages/dm-core-plugins/src/view_selector/TabsPlugin.tsx
+++ b/packages/dm-core-plugins/src/view_selector/TabsPlugin.tsx
@@ -1,4 +1,4 @@
-import { IUIPlugin, Loading } from '@development-framework/dm-core'
+import { IUIPlugin } from '@development-framework/dm-core'
 import { TViewSelectorConfig } from './types'
 import * as React from 'react'
 import { Tabs } from './Tabs'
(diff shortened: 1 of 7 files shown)


packages/dm-core-plugins/src/view_selector/TabsPlugin.tsx
```diff
@@ -26,7 +26,7 @@ export const TabsPlugin = (
     throw new Error(JSON.stringify(error, null, 2))
   }
   if (isLoading || !viewSelectorItems.length || !selectedViewId) {
-    return <Loading />
```
henrikbossart, line 29: Why remove the loading animation?

henrikbossart (approved):
Looks good! Just one question.

packages/dm-core-plugins/src/view_selector/TabsPlugin.tsx
eoaksnes: To avoid too many loading signs pop-up/flickering. And just make the plugins responsible for showing the loading sign.

PR #687, opened 2023-11-02
feat: expandable job log

henrikbossart:
Adjusts the look of the logs, results and errors of jobs to be more like highlighted code-blocks. Also makes it possible to open them in a dialog, to get a larger box for viewing them.

I'm not too sure about the format of logs, as the type is just a string. That makes it a bit hard to plan for the ui.

Some screenshots:
Screenshot 2023-11-02 at 07 56 29
Screenshot 2023-11-02 at 07 56 33


henrikbossart:
Visualization with some more realistic logs:
Screenshot 2023-11-02 at 08 48 03
Screenshot 2023-11-02 at 08 48 07

packages/dm-core-plugins/src/job/JobPlugin.tsx
```diff
@@ -188,30 +235,86 @@ export const JobPlugin = (props: IUIPlugin & { config: JobPluginConfig }) => {
...
+              <Typography variant="h6" style={{ paddingBottom: '.4rem' }}>
+                Logs:
+              </Typography>
+              {/*<Button variant="ghost_icon" size={16} label='Expand logs' onClick={() => setShowLogDialog(true)}><Icon*/}
+              {/*  data={expand_screen}/></Button>*/}
+            </div>
+            <FormattedLogContainer
+              style={{ maxWidth: '40rem', maxHeight: '20rem' }}
+            >
+              <code
```
soofstad, line 278: Seems to be a bit of duplicated code between this and the "JobsLogsDialog", possible to share some of it?

packages/dm-core-plugins/src/job/JobLogsDialog.tsx
```diff
@@ -0,0 +1,88 @@
...
+                        language: 'json',
+                      }).value,
+                    }
+                  : {
+                      __html: hljs.highlightAuto(logs).value,
+                    }
+              }
+            />
+          </FormattedLogContainer>
+        </div>
```
soofstad, line 63: This could be a bad idea.
We have no controll of what's being logged. Malicous requests could be logged, and combined with this, will lead to a XSS vulnerability.

Also, logs are often not JSON, so this highlighting is not that critical, IMO :)

soofstad (changes requested):
Looks much better :+1: 
A few comments

packages/dm-core-plugins/src/job/JobPlugin.tsx
henrikbossart: yeah, can probably be extracted to a common place, I just wanted to know about the general direction 👍

packages/dm-core-plugins/src/job/JobLogsDialog.tsx
henrikbossart: indeed.
We can use `dompurify` to sanitize the logs, and prevent ( at least to a higher degree) XXS attacks

henrikbossart: oh, and about the hightlighting; I thought that was something that was mentioned in a meeting or something? that it was desirable to make it look slightly like the docker logs?

henrikbossart:
@soofstad I changed the log renderer to only highlight if the content is an actual object.

henrikbossart:
I've now removed all the highlighting from the job log plugin, so that it is only white text on a dark backdrop.

I've also chosen to sanitize the html injection happening in the Yaml plugin.

soofstad (approved):
Good stuff! :closed_lock_with_key:

PR #696, opened 2023-11-07
fix(form): add or remove optional lists

eoaksnes:

* fix: add or remove optional lists
* fix: tab crashes if try to open a list that does not exists (need to add it first)
* refactor: moved `AddObjectButton` and `RemoveObjectButton` into their own files and out from `ObjectField` that was getting to big, and also re-using the buttons in the `ArrayField`.
* refactor: split up `ArrayField` which was too complex, into using several components `PrimitiveList`, `InlineList`, or `OpenList`, instead of everything being one huge component. 
* chore: update DMSS version that includes fix for add optional empty lists 

How it looks like:

image

UiRecipes exist, so can be removed or opened.

StorageRecipes does not exist, so can add it.

### DIFF
--- a/example/docker-compose.yaml
+++ b/example/docker-compose.yaml
@@ -2,7 +2,7 @@ version: '3.4'
 
 services:
   dmss:
-    image: datamodelingtool.azurecr.io/dmss:v1.7.1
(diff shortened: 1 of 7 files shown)


pbullhove (approved):
sweeet

PR #697, opened 2023-11-07
chore: remove unused env

eoaksnes:

* REACT_APP_DM_JOB_URL is not in use anymore, remove it from startup script and at the same time the `cross-env` dependency 💥


soofstad (approved):
:tada:

henrikbossart (approved):

PR #704, opened 2023-11-09
test: examples using default dmss ui recipe

### LINKED ISSUE #599: test: New test that checks rendering of default_ui_recipes
Goal is to test that it is possible to render entities where there is no special Recipe given.
That means the type gets its uiRecipes from DMSS default_ui_recipe

roar-larsen:
Added examples using the default DMSS Ui Recipe where there are no dedicated ui recipes.

closes #599

### DIFF
--- a/e2e/tests/plugin-default_recipe.spec.ts
+++ b/e2e/tests/plugin-default_recipe.spec.ts
@@ -1,94 +1,33 @@
 import { expect, test } from '@playwright/test'
 
 test('Form default DMSS UI Recipe', async ({ page }) => {
-  const navigate = async () => {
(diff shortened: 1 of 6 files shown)


e2e/tests/plugin-default_recipe.spec.ts
soofstad, line 1: Perhaps don't have this test every aspect of the form plugin (other tests do that), but rather just check that the default recipe renders as expected.

Having too many similar tests only create lots of extra work when something changes, without much benefits.

soofstad (changes requested):

soofstad (approved):
:+1:

PR #707, opened 2023-11-10
refactor: lazy import plugins

eoaksnes:

* Lazy load plugins, see https://legacy.reactjs.org/docs/code-splitting.html

### DIFF
--- a/packages/dm-core-plugins/src/index.tsx
+++ b/packages/dm-core-plugins/src/index.tsx
@@ -1,71 +1,108 @@
 import { TUiPluginMap } from '@development-framework/dm-core'
 
-import { YamlPlugin } from './yaml/YamlPlugin'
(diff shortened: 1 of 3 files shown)


henrikbossart (approved):

packages/dm-core-plugins/src/index.tsx
```diff
@@ -1,71 +1,108 @@
...
   '@development-framework/dm-core-plugins/explorer': {
-    component: ExplorerPlugin,
+    component: lazy(() => import('./explorer/ExplorerPlugin')),
   },
   '@development-framework/dm-core-plugins/list': {
-    component: ListPlugin,
+    component: lazy(() =>
+      import('./list/ListPlugin').then((module) => ({
+        default: module.ListPlugin,
+      }))
```
soofstad, line 15: Very cool :D 
Why  is the syntax different in these plugins compared to "explorer"?
Looks different from the react docs as well

soofstad (approved):

PR #708, opened 2023-11-10
refactor: tweak the user info dialog

henrikbossart:
Slightly tweaks the user info dialog

Screenshot 2023-11-10 at 15 31 27
Screenshot 2023-11-10 at 15 31 30

### DIFF
--- a/packages/dm-core-plugins/src/header/components/UserInfoDialog.tsx
+++ b/packages/dm-core-plugins/src/header/components/UserInfoDialog.tsx
@@ -1,7 +1,7 @@
 import { Dialog, RoleContext, useDMSS } from '@development-framework/dm-core'
 import { Button, Radio, Typography } from '@equinor/eds-core-react'
 import { AxiosResponse } from 'axios'
-import React, { useContext, useRef, useState } from 'react'
+import React, { useContext, useState } from 'react'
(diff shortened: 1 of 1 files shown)


packages/dm-core-plugins/src/header/components/UserInfoDialog.tsx
```diff
@@ -82,33 +89,43 @@ export const UserInfoDialog = (props: UserInfoDialogProps) => {
...
+            >
+              Create API-Key
+            </Button>
+          </FlexRow>
+          <FlexRow>
+            <Button variant="ghost" color="danger" onClick={() => logOut()}>
+              Log out
+            </Button>
+            <Button onClick={() => setIsOpen(false)}>
+              {selectedRole !== originalRole.current ? 'Save' : 'Cancel'}
```
soofstad, line 125: See pitfalls of useRef: https://react.dev/reference/react/useRef
 "Don't read a ref during rendering"
 Also, this is still a bit misleading UI. Perhaps move the "setSelectedRole" to clicking the "Save" button. Then this ref wont be needed.

henrikbossart: right, my brain is still coupled to a different way of thinking about `ref`

soofstad (approved):

PR #732, opened 2023-11-14
fix(form): recipe name should be optional in ObjectField

### LINKED ISSUE #718: BUG: Duplicate UI recipe issue

soofstad:
- "form"-plugin will now default to use the first uiRecipe in the list of recipes (including "initial")
- If a recipe name was given, but not found --> Raise error.

Fixes a bug where "form" failed to pick a uiRecipe if no name was provided

closes #718


henrikbossart (approved):
LGTM 👍

PR #737, opened 2023-11-15
feat: refresh button

pbullhove:

Adding refresh button

### DIFF
--- a/packages/dm-core/src/components/EntityView.tsx
+++ b/packages/dm-core/src/components/EntityView.tsx
@@ -1,4 +1,4 @@
-import React, { Suspense } from 'react'
+import React, { useState, Suspense } from 'react'
 
 import styled from 'styled-components'
 import { ErrorBoundary, ErrorGroup } from '../utils/ErrorBoundary'
@@ -6,11 +6,13 @@ import { useRecipe } from '../hooks'
 import { IUIPlugin } from '../types'
 import { Loading } from './Loading'
 import { Typography } from '@equinor/eds-core-react'
+import RefreshButton from './RefreshButton'
 
 const Wrapper = styled.div`
   align-self: start;
   justify-content: space-evenly;
   width: 100%;
+  position: relative;
 `
 
 type IEntityView = IUIPlugin & {
@@ -28,13 +30,19 @@ export const EntityView = (props: IEntityView): React.ReactElement => {
     dimensions
   )
 
+  // Refresh Button stuff
+  const [reloadCounter, setReloadCounter] = useState(0)
+  const [hoverOver, setHoverOver] = useState({
+    refreshButton: false,
+    component: false,
+  })
+
   if (isLoading)
     return (
       <div style={{ alignSelf: 'center', padding: '50px' }}>
         <Loading />
       </div>
     )
-
   if (error)
     return (
       <ErrorGroup>
@@ -44,7 +52,6 @@ export const EntityView = (props: IEntityView): React.ReactElement => {
         <pre>{JSON.stringify(error, null, 2)}</pre>
       </ErrorGroup>
     )
-
   if (!recipe || !Object.keys(recipe).length)
     return <Wrapper>No compatible uiRecipes for entity</Wrapper>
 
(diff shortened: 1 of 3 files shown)


henrikbossart:
Looks good, need to support configuring the button in the entity though.
Only gripe I have is that is looks slightly scuffed with the lack of space/padding between the border and the content when hovering. Not quite sure how to fix that, as it would need some padding on the content. And I don't think that works for all cases

packages/dm-core/src/components/EntityView.tsx
```diff
@@ -28,13 +37,29 @@ export const EntityView = (props: IEntityView): React.ReactElement => {
     dimensions
   )
 
+  // Refresh Button stuff
+  const [reloadCounter, setReloadCounter] = useState(0)
+  const [mouseHoverOverRefreshButton, setMouseHoverOverRefreshButton] =
+    useState(false)
+  const [mouseCloseToRefreshButton, setMouseCloseToRefreshButton] =
+    useState(false)
```
soofstad, line 45: Cannot this hover effect be more elegantly be handled with CSS?
I think so... :)

packages/dm-core/src/components/EntityView.tsx
```diff
@@ -28,13 +37,29 @@ export const EntityView = (props: IEntityView): React.ReactElement => {
...
+  const reloadComponent = () => setReloadCounter(reloadCounter + 1)
```
soofstad, line 46: Not a big fan of these functions that only call another function. IMO it's better to call "setReloadCounter" directly.
Helps with readability

soofstad (changes requested):

soofstad:
Looks good. Works great.
Think we can wait with making it configurable.
Also, move the button just a few pixles left and down, so that it does not overlap with the border?
And, is there something we can do to make it more discoverable. It's quite hard to find the button in some of the plugins/grids/layout. Perhaps make the icon itself always be visible, with high opacity. Experiment and see what works :shrug:

henrikbossart:
>  Perhaps make the icon itself always be visible, with high opacity.

I assume 'low opacity' is meant?

packages/dm-core/src/components/EntityView.tsx
pbullhove: Hmm how so?

By wrapping it in a invisible container of 1/3 size and having the mouseEnter and mouseLeave property here?

packages/dm-core/src/components/EntityView.tsx
pbullhove: alright

packages/dm-core/src/components/EntityView.tsx
henrikbossart: Could probably make the button opacity work with only css, but I don't think you can style different elements based on hovering another with pure CSS 🤔

pbullhove:
> Looks good, need to support configuring the button in the entity though. Only gripe I have is that is looks slightly scuffed with the lack of space/padding between the border and the content when hovering. Not quite sure how to fix that, as it would need some padding on the content. And I don't think that works for all cases

Managed to fix this with 'outline' instead of 'border' which does not _shift_ the content inside, and by adding a property 'outlineOffset'.

packages/dm-core/src/components/EntityView.tsx
soofstad: you can set different styling based on "hover" and other similar properties like so:
```
const ClickableIcon = styled.button`
  appearance: none;
  border: none;
  background-color: transparent;

  &:hover {
    color: gray;
    cursor: pointer;
  }
`
```

henrikbossart:
> > Looks good, need to support configuring the button in the entity though. Only gripe I have is that is looks slightly scuffed with the lack of space/padding between the border and the content when hovering. Not quite sure how to fix that, as it would need some padding on the content. And I don't think that works for all cases
> 
> Managed to fix this with 'outline' instead of 'border' which does not _shift_ the content inside, and by adding a property 'outlineOffset'.

Right, but doesn't the outline give a rather ugly "edgy" outline? 🤔

pbullhove:
> > > Looks good, need to support configuring the button in the entity though. Only gripe I have is that is looks slightly scuffed with the lack of space/padding between the border and the content when hovering. Not quite sure how to fix that, as it would need some padding on the content. And I don't think that works for all cases
> > 
> > 
> > Managed to fix this with 'outline' instead of 'border' which does not _shift_ the content inside, and by adding a property 'outlineOffset'.
> 
> Right, but doesn't the outline give a rather ugly "edgy" outline? 🤔 Or does it curve smoothly like the border?

Can have border-radius on the outline as well.

example/app/data/DemoDataSource/recipes/plugins/grid/car_grid/car.recipe.json
```diff
@@ -3,7 +3,7 @@
   "_blueprintPath_": "/plugins/grid/car_grid/blueprints/Car",
   "initialUiRecipe": {
     "name": "form",
-    "type": "CORE:UiRecipe",
+    "type": "CORE:git ",
```
henrikbossart, line 6: this seems like an unintended change?

pbullhove: hhehe nice catch

soofstad (approved):
Looks great!

PR #746, opened 2023-11-17
fix: resolve relative references

eoaksnes:

Fixes the problem in SIMPOS where cant view signals:

image


pbullhove (approved):

soofstad (approved):
Is this a new issue? Or does this fix the problem with resolving "^"-addresses in a list?

PR #747, opened 2023-11-17
feat(form): show if a value has changed

eoaksnes:

Show if a value has changed:

image


soofstad (changes requested):
IMO, the "warning" variant looks like a validation error. And I don't think we should use it for dirty state.

Perhaps a small symbol as "rightAdornment" is better suited for indicating dirty state?

pbullhove (approved):
Very clear. 
Does it make sense to have something like this for the objectField and Arrayfield as well, or are they never in dirty-state?

henrikbossart:
I agree with @soofstad here, if you are going for a color, use a more neutral one.

PR #748, opened 2023-11-18
feat(form): configure if fields should be expanded by default

eoaksnes:

Add config to say if inline objects or arrays should be collapsable/expanded by default or not!

image

image


pbullhove (approved):
Nice. 
Only comment I have is that perhaps the blueprintAttribute was repeated a bunch of times, perhaps have that as a root blueprint and inherit from that?

soofstad (approved):

PR #749, opened 2023-11-20
chore: bump dmss and dm-job versions

roar-larsen:
Bump dmss and dm-job versions

### DIFF
--- a/example/docker-compose.yaml
+++ b/example/docker-compose.yaml
@@ -2,7 +2,7 @@ version: '3.4'
 
 services:
   dmss:
-    image: datamodelingtool.azurecr.io/dmss:v1.8.0
+    image: datamodelingtool.azurecr.io/dmss:v1.10.0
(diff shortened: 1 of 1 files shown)


pbullhove (approved):

example/docker-compose.yaml
```diff
@@ -2,7 +2,7 @@ version: '3.4'
 
 services:
   dmss:
-    image: datamodelingtool.azurecr.io/dmss:v1.8.0
+    image: datamodelingtool.azurecr.io/dmss:v1.10.0
```
soofstad, line 5: might as well set these to "latest"

soofstad (approved):

example/docker-compose.yaml
roar-larsen: We had a discussion on this earlier this summer (when you were away) and decided then not to use latest anymore. But we could revisit this.

PR #758, opened 2023-11-20
feat: list clickable and chevron change

pbullhove:

1. Chevron is now right and down, instead of down and up. This is to be consistent with the other usages of chevron in the application. Like how it is used in the tree view. 
image
2. List title is now clickable as well, not only the chevron. Makes it easier to use.


eoaksnes (approved):
👍 Nice

PR #763, opened 2023-11-21
refactor: move media content component to dm-core

henrikbossart:
Moves the `MediaContnet` component into dm-core, as its needed in both the `MediaViewerPlugin` and in the report in simpos.

This PR is needed for equinor/dm-app-simpos#51 to work.


packages/dm-core/src/components/MediaContent.tsx
```diff
@@ -102,7 +107,7 @@ export function MediaContent(props: MediaContentProps): ReactElement {
             <label className="meta-label">Filetype:</label>
             <span> {meta.filetype}</span>
             <label className="meta-label">Filesize:</label>
-            <span> {formatBytes(meta.fileSize)}</span>
+            {/*<span> {formatBytes(meta.fileSize)}</span>*/}
```
soofstad, line 110: Why remove this?

soofstad (approved):

packages/dm-core/src/components/MediaContent.tsx
henrikbossart: oh, Thats not intended removed! thanks!

henrikbossart: Had to move that function over to `dm-core` though..

PR #785, opened 2023-11-22
fix(Sidebar): parnet menu items shouldn't be clickable

soofstad:
Make parent menu entries in Sidebar™ not be clickable.


henrikbossart:
why should they not be clickable? 🤔

soofstad:
It messes with the navigation. Makes stuff jump around when you are just exploring the menus.
Also, requested by user representative :)

henrikbossart (approved):

PR #786, opened 2023-11-22
feat: tailwindcss

henrikbossart:
Adds tailwindcss to the project.

its sooo cooool and awesome and mega mega convenient! 😍 

Note, its nessecary to run `yarn packages:build` for tailwind to build in the packages, and for the `build.css` file to become available


pbullhove (approved):
className={"lovely-PR"}

henrikbossart:
OK, so I've changed things a bit.

Now, tailwind is added to both `dm-core` and `dm-core-packages`, and when the packages are built, the tailwind "compiler" generates a `build.css` (minified) file, which is copied to the dist directory. This means that when other users are importing our plugins, the styling should still work.

eoaksnes (approved):
💯 smooth

PR #798, opened 2023-11-24
Perf: run playwright parallel

roar-larsen:
Changed from 1 to 4 workers. Run time is reduced approx 1 minute.

Tried with 6 workers also, but it kept failing. Need to monitor if 4 workers is stable enough.

### DIFF
--- a/.github/workflows/integration-tests.yaml
+++ b/.github/workflows/integration-tests.yaml
@@ -17,7 +17,7 @@ on:
 jobs:
   integration-tests:
     timeout-minutes: 60
-    runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


soofstad:
Think we should test with a bigger runner for more workers before merging this.
I've contacted Toolbox team for help with access to a larger runner

soofstad (approved):

roar-larsen:
Done some testing with a bigger runner and it seems to be somewhat faster (from almost 3 min, down to 2 min). Runs fine with both 4 and 6 workers, but there is no difference in performance between 4 and 6 workers.

PR #800, opened 2023-11-24
feat: table plugin improvements

awesthouse:
- Adds helpers component ConditionalWrapperto dm-core
- Adds components SortableList and SortableListItem to dm-core.
    - By wrapping elements in the Sortable components you should be able to use drag and drop without much configuration at all
- Refactors and expands Table plugin config
- Adds component Table to dm-core
  - Two variants exist: view and edit. You can configure which ones to show (one or both, both gives user oppertunity to toggle between). You can also configure functionality available on each individual variant.
  - View variant functionality: Sort by column, add, delete, open in tab, open as expandable
  - Edit variant: functionality: Sort items using drag and drop, add, delete. edit values

Expanding already existing Table plugin

#258


collinlokken (commented):
I checked out but was not able to run, I get the same error that the integration tests get. 
image

packages/dm-core-plugins/src/table/utils.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export function mergeConfigs(config: TTableConfig): TTableConfig {
+  return {
+    ...defaultConfig,
+    ...config,
+    variant: config.variant.map((variant) => ({
+      ...variant,
+      functionality: {
+        delete: variant.functionality?.delete !== false ? true : false,
+        add: variant.functionality?.add !== false ? true : false,
```
henrikbossart, line 42: This can be simplified:
```suggestion
        delete: variant.functionality?.delete !== false,
        add: variant.functionality?.add !== false,
```

henrikbossart (changes requested):
The dnd package is adding a `div` element inside the `tbody`, which isn't allowed, this causes an error to be thrown in the console. Maybe it could be changed to a `<template>` tag, or moved out of the table?

packages/dm-core-plugins/src/table/utils.ts
awesthouse: Ah, nice! Thank you 👍🏻 This didn't even look right when I wrote it, but my brain had no intention of working

awesthouse:
> The dnd package is adding a `div` element inside the `tbody`, which isn't allowed, this causes an error to be thrown in the console. Maybe it could be changed to a `<template>` tag, or moved out of the table?
> 
> Other than that this looks very good! Only two small nitpicks :)

Thank you! After doing some digging - realized that the div came from dnd context, not the sortable list context that needs to only wrap the sortable elements so by separating them we can actually move the larger dnd context out of the table 👍🏻 I think it makes sense anyways 'cause you can nest dnd contexts as well so it does make sense that it is it's own component

eoaksnes (approved):

PR #805, opened 2023-11-24
fix: generate new job api and update job plugin new api

collinlokken:
- Generates new job api based on the new endpoint for updating progress
- Introduces progress bar to jobs

Job API client needs to be updated to reflect the changes in dm-job


pbullhove (approved):
Good work

PR #807, opened 2023-11-25
refactor: avoid prop drilling

eoaksnes:

* Avoid prop drilling, only pass what is needed
  * Use context to provide the `FormConfig` to fields, instead of passing them down as props (the read only and show expanded setting)
  * Add util `getDisplayLabel` to provide a way to have common labels for all fields, instead of passing it down as a prop
  * Just pass the `attribute` down instead of splitting it up and passing down parts of it
* Make `StringField` less complex by
  * Add `NumberWidget` to handle numbers
  * Add  `DateTimeWidget` to handle date time

> **NOTE**: The overall goal of this PR is to make it clear what props are needed for each field (StringField, NumberField etc.). This PR focuses mostly on primitive fields.  The PR is quite large already, since we are renaming and removing stuff, so will not include everything in this PR, more to come....

### DIFF
--- a/e2e/tests/plugin-form-Nested.spec.ts
+++ b/e2e/tests/plugin-form-Nested.spec.ts
@@ -35,7 +35,7 @@ test('Change owner', async () => {
     .fill('Jacob')
   await page
     .getByRole('tabpanel')
-    .getByTestId('form-text-widget-Phone Number (optional)')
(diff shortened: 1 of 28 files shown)


collinlokken (approved):
Nice 👍

PR #808, opened 2023-11-26
refactor: add templates to divide code more up

eoaksnes:

* Add templates to divide the code more up and more easy to understand/maintaine

image

> **NOTE**: This PR is based on #807, so that needs to be taken in first.


collinlokken (approved):
👍

PR #812, opened 2023-11-28
fix: simple job control button

collinlokken:
- Turns the job controll button (run/stop/restart) into a simple button with only one icon
- Do stop and create in two steps

### DIFF
--- a/packages/dm-core-plugins/src/job/JobControlButton.tsx
+++ b/packages/dm-core-plugins/src/job/JobControlButton.tsx
@@ -1,7 +1,7 @@
 import { Button, CircularProgress, Icon } from '@equinor/eds-core-react'
-import { play, stop, refresh } from '@equinor/eds-icons'
+import { play, stop, refresh, IconData, calendar } from '@equinor/eds-icons'
 import { JobStatus } from '@development-framework/dm-core'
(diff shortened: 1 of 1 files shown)


collinlokken:
image (1)
image (2)
image (3)
image (4)
image

packages/dm-core-plugins/src/job/JobControlButton.tsx
```diff
@@ -13,57 +13,70 @@ export const JobControlButton = (props: {
 }) => {
   const { jobStatus, createJob, asCronJob, disabled, exists, remove } = props
   const [hovering, setHovering] = useState(false)
+  const [buttonColor, setButtonColor] = useState<
+    'primary' | 'secondary' | 'danger' | undefined
```
henrikbossart, line 17: Why would the color ever be undefined? 🤔

collinlokken: Does not really make sense, removed it 👍

henrikbossart (approved):
The red danger color is _very_ red, but I guess its ok... 🤷‍♂️

soofstad:
Looks good!
The red color should be the same red color.

henrikbossart:
@soofstad I think the red is the same, just in hover state and not

PR #813, opened 2023-11-28
feat: update state of parent form on changes in nested forms

eoaksnes:

If `onSubmit` is passed into the form plugin remove the submit button, and instead call the `onSubmit` on any changes to notify any parent plugins of changes.

> **NOTE**:  This is only for inline forms that are model-contained


soofstad (approved):
Looks good!

PR #826, opened 2023-11-29
feat: replace eslint and prettier with biome

henrikbossart:
Replaces ESlint and prettier with a new, faster toolchain called [biome](http://biomejs.dev)
Also adds `incremental: true` to the `tsconfig` to speed up the typechecking.

Biome is a new tool that is built in rust, and is a lot faster than the old tools. I thought I would try it out, with the goal of reducing the time it takes to check the project before each commit. It does reduce the time, but not as much as I hoped.

One thing it did do, is point out how many rules we are ignoring (see the `"rules"` in `biome.json`. It also formatted the entire project to the same standard, since we now only have one config for everything. Its still possible to have "local" configs that extend the base though.

The only question now is tabs, or spaces....

There are extensions for biome in both VSCode and IntelliJ

(If we actually want to do this switch, we'll have to make some changes to our CI)


soofstad:
Anything that get's rid of eslint get's my vote :smile: 
However, 8 spaces indentation seems a bit excessive

PR #827, opened 2023-11-29
Feat/dimentional skalar primitives

pbullhove:


henrikbossart (approved):

eoaksnes (approved):
👍 smooth

packages/dm-core-plugins/src/form/fields/NumberField.tsx
```diff
@@ -23,23 +23,23 @@ export const NumberField = (props: TField) => {
       }}
       defaultValue={attribute.default || 0}
       render={({
-        field: { ref, onChange, ...props },
+        field: { ref, value, onChange, ...props },
```
soofstad, line 26:
```suggestion
        field: { ref, value, onChange },
```

packages/dm-core-plugins/src/form/widgets/DimensionalScalarWidget.tsx
```diff
@@ -33,7 +38,7 @@ const DimensionalScalarWidget = ({
       <TextField
         unit={widgetConfig?.unit || entity?.unit}
         id={entity?.id}
-        type="number"
+        type={isNumber ? 'number' : undefined}
```
soofstad, line 41: Does it even make sense that a scalar can be anything besides "Number"? :thinking:

packages/dm-core-plugins/src/form/widgets/DimensionalScalarWidget.tsx
```diff
@@ -45,11 +50,18 @@ const DimensionalScalarWidget = ({
         }
         disabled={readOnly}
         onChange={(e: ChangeEvent<HTMLInputElement>) => {
-          if (e.target.value === '') setError('Value is required')
-          else setError('')
-          onChange({ ...entity, value: parseFloat(e.target.value) })
+          setError(e.target.value === '' ? 'Value is required' : '')
```
soofstad, line 53: Does this make every scalar required?
Thats probably not right

soofstad (changes requested):

packages/dm-core-plugins/src/form/widgets/DimensionalScalarWidget.tsx
pbullhove: Could be both strings and numbers, even though the name "scalar" implies a number, there is no need to enforce the type of the data to be number and not string. 
Works just as well as a text box.

packages/dm-core-plugins/src/form/fields/NumberField.tsx
pbullhove: Ok

packages/dm-core-plugins/src/form/widgets/DimensionalScalarWidget.tsx
pbullhove: discussed live

soofstad (approved):

PR #828, opened 2023-11-29
fix: model uncontained and media_viewer tests

roar-larsen:
Added some extra asserts and other changes to try to make the test more stabile.

If the media_viewer test still is flaky, we could try to use expect.toPass to retry the assertions. The reason the test was flaky seem to be that the test proceeded before the image was completely loaded.

Tests were flaky.


pbullhove (approved):
Good!

PR #833, opened 2023-11-29
Fix/media viewer bugs

henrikbossart:
So chrome does in fact have controls for pdfs, the iframe was just too small. Increasing the size of the iframe solved the issue :+1:

What I'm not so sure how to solve is the fact that chrome downloads videos, instead of showing them in a tab, and it doesn't seem to be possible to mitigate this behaviour without installing some chrome extension... Could conditionally show the tab button based on browser, but I'm not sure how ideal that is.....

Also changes some header things, becuase the selector thingy was very small.


soofstad (commented):
Nice!

Don't worry too much about the video thing.
Videos most likely needs to be either streamed or downloaded anyway. Not many videos will be suitable for full loading in the browser.

soofstad (approved):

PR #839, opened 2023-11-30
feat: partial update

eoaksnes:

* Add support to the form plugin so that it only sends a payload to the API with attributes that should be updated (are visible)


collinlokken (approved):
👍

PR #842, opened 2023-11-30
fix: model uncontained failing in ci

roar-larsen:
Added another assert to basically slow down the test.

test keeps failing randomly in ci.


soofstad (approved):

roar-larsen:
Closing PR. Trying to fix through #846 instead.

PR #854, opened 2023-12-01
Feat/role mapping

soofstad:
NOTE: Depends on https://github.com/equinor/data-modelling-storage-service/pull/713

- Adds "roles" to the application config
- Update "RoleContext" to use these roles, and generate fallbacks if missing

closes https://github.com/equinor/data-modelling-storage-service/issues/695


eoaksnes (approved):
👍 smooth

PR #870, opened 2023-12-04
refactor: use new datepicker

henrikbossart:
Replaces the old datepicker with the new datepicker.
Also makes some changes to how the props and interface of the datepicker, to make it fit more with the previous signature.


eoaksnes (approved):
👍

PR #871, opened 2023-12-04
feat: better blueprint plugin

pbullhove:

Modifies the old blueprint plugin. 

1. Better UI
2. Fixes bug of restoring that did not restore everything

image

### DIFF
--- a/packages/dm-core-plugins/src/blueprint/BlueprintPlugin.tsx
+++ b/packages/dm-core-plugins/src/blueprint/BlueprintPlugin.tsx
@@ -1,224 +1,33 @@
 import {
-  BlueprintPicker,
-  EPrimitiveTypes,
   INPUT_FIELD_WIDTH,
   IUIPlugin,
   Loading,
-  Select,
+  TAttribute,
   TBlueprint,
-  TGenericObject,
-  truncatePathString,
   useDocument,
 } from '@development-framework/dm-core'
 import * as React from 'react'
-import { ChangeEvent, useEffect, useState } from 'react'
-import {
-  Accordion,
-  Button,
-  Icon,
-  Input,
-  Label,
-  Switch,
-  TextField,
-} from '@equinor/eds-core-react'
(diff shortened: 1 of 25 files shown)


packages/dm-core-plugins/src/blueprint/BlueprintPlugin.tsx
```diff
@@ -1,231 +1,28 @@
...
-    </div>
-  )
-}
-
 export const BlueprintPlugin = (props: IUIPlugin) => {
   const { idReference } = props
   const { document, isLoading, updateDocument, error } =
     useDocument<TBlueprint>(idReference)
   const [formData, setFormData] = useState<any>({ ...document }) //TODO remove any type (requires TBlueprint to be updated)
+  const dirtyState = !isEqual(document, formData)
```
soofstad, line 25: Would a "useMemo" help on performance here?

soofstad (approved):
Looks good! :100:

packages/dm-core/src/components/Pickers/BlueprintPicker.tsx
```diff
@@ -33,15 +35,14 @@ export type TBlueprintPickerProps = {
   /** A title for the picker */
   label?: string
   /** Render a input element or just a button */
-  type?: 'button' | 'input'
+  type?: 'button'
```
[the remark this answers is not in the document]
pbullhove: It was not used. 

Also, it behaved really weirdly. It was an input-box that when you clicked it behave just like the button, opening the tree-view. 
So then we can just use the button.

packages/dm-core-plugins/src/blueprint/BlueprintPlugin.tsx
pbullhove: yes, very much! Thank you for the pointer.

henrikbossart:
I think there's a merge that's gone wrong here :/

PR #872, opened 2023-12-04
feat: update items if changed in expanded or tabs

eoaksnes:

* Added `onChange` to the UiPlugin interface that can be used to listen to changes in child plugins. 

Updated form behavior:

* If the `onChange` is passed to the `FormPlugin` it will hide the submit button and instead call the `onChange` on any updates. 
* If the `onSubmit` is passed to the `FormPlugin` it will call `onSubmit` if the document is submitted.  

Added extra configuration to the `ListPlugin` to control behavior. Added `saveExpanded` (default is false) to the `ListPluginConfig` that controls the saving of expanded items. 
  * If `saveExpanded` is set to true, then the `ListPlugin` listens to changes (onChange) of expanded items and it will control the save.
  * If `saveExpanded` is set to false, then the `ListPlugin` listens to submits (onSubmit) of expanded items and only updates display information.

Other changes:

* The `ListPlugin` will listen to `onSubmit` for items open in tabs, to be able to update itself with correct information.
* The `FormPlugin` will listen to `onChange` for nested forms, so that the root form should show the submit button and control the save.


collinlokken (approved):
nice!

PR #892, opened 2023-12-05
feat: Removes up-down arrows in numberfields

pbullhove:

No arrows, looks nicer. 

image

#893


henrikbossart:
I'm not going to speak too loudly on this, but isn't this a very bad move in regards of user friendlieness and user experience? Wouldn't it be better to replace the arrows with nicer arrows?

care to chime in on the UX part of this, @awesthouse ?

pbullhove:
@henrikbossart I don't think so.. 
I mean this is a completely normal way to imput text, so why should numbers be so hard without the arrows?

Excel does not have arrows..

henrikbossart (approved):

PR #901, opened 2023-12-06
fix: list clicking and defaults

peter-ryen-bull:

1. Makes list title clicking expand item
2. Restores list defaults to be "open": false


soofstad (approved):
:partying_face:

PR #902, opened 2023-12-06
feat: move softbutton into core and use it in table as well

pbullhove:


packages/dm-core/src/components/Table/TableRow/TableRowActions/.TableRowActions.tsx.swp
henrikbossart, line 1: Whats this file? 🤔

henrikbossart (approved):

PR #904, opened 2023-12-06
Feat/hard delete button

pbullhove:

Adds a delete button with pop-up dialogue that required confirmation on delete. 
Can be used on mulitple deleting stuff. Must have a wrapper that defines what happens onConfirmDelete

- Is an x instead of the old trash can
- Has a tooltip with "delete + namePath". 

image
image
image


eoaksnes:
https://medium.com/@wdeamonica/x-to-delete-or-x-for-error-misunderstanding-x-icon-in-user-interface-1fd57baa2f5b

pbullhove:
I think it's a valid point, that in this context the x is less intuitive. 
It might also signify harder "Delete whole attribute". 

I will go through form today, and look at how to make it more clear. 
I think moving the X to the right corner instead of so close to the attribute might make it much more clear that it's a delete.

PR #907, opened 2023-12-06
feat: add scope to columns

eoaksnes:

The `Engine` is a complex attribute, and here we just show the name of it in the table.

image

> **NOTE**: The edit mode also works on scope. PS! For the engine, it, for now, needs to be created first in expanded view since it is optional and the add row does not create it by default. I think more logic should be added so that the engine is created automatically. I did not want to add all this extra stuff in one PR, so I just kept it simple.


packages/dm-core/src/components/Table/utils.ts
```diff
@@ -15,8 +31,12 @@ export function updateItemAttribute(
 ) {
   const itemsCopy = [...items]
```
soofstad, line 32: Sure we don't want a deep copy with "structuredClone" here?

packages/dm-core/src/components/Table/utils.ts
```diff
@@ -15,8 +31,12 @@ export function updateItemAttribute(
...
   const index = itemsCopy.findIndex((item) => item.key === key)
-  if (itemsCopy[index].data) {
-    ;(itemsCopy[index].data as TGenericObject)[attribute] = newValue
+  const itemData = itemsCopy[index].data || {}
+  const success = setValue(itemData, attribute, newValue)
+  if (success) {
+    if (itemsCopy[index].data) {
+      ;(itemsCopy[index].data as TGenericObject) = itemData
+    }
```
soofstad, line 39:
```suggestion
    if (success && itemsCopy[index].data) {
      ;(itemsCopy[index].data as TGenericObject) = itemData
    }
```

soofstad (approved):

PR #920, opened 2023-12-07
chore: only show progress bar when job is running

henrikbossart:
as title says


soofstad (commented):
Change is ok.
But if we are refining this, perhaps only show progress if running and progress is not "None". Since not all jobs will report progress, so showing the progress bar will be missleading

soofstad (approved):
Guuuud

PR #922, opened 2023-12-07
Feat: model uncontained form UI 

pbullhove:

image


packages/dm-core-plugins/src/form/widgets/TextareaWidget.tsx
```diff
@@ -14,13 +14,18 @@ const TextareaWidget = (props: TWidget) => {
 
   return (
     // @ts-ignore
```
henrikbossart, line 16:
```suggestion
```

packages/dm-core-plugins/src/form/widgets/TextareaWidget.tsx
```diff
@@ -14,13 +14,18 @@ const TextareaWidget = (props: TWidget) => {
...
-      {...props}
-      multiline={true}
-      rows={5}
-      onChange={onChangeHandler}
-      label={label}
-    />
+    <>
+      <Typography className={'self-center pb-2'} bold={true}>
+        {label}
+      </Typography>
```
henrikbossart, line 20: If you wrap this in a `<label htmlFor={label}>` it will tell the screen-readers about which field it belongs to. ({label} because thats whats used as id in the `TextField` component)

packages/dm-core-plugins/src/form/widgets/TextWidget.tsx
```diff
@@ -9,26 +9,30 @@ const TextWidget = (props: TWidget) => {
...
-              // @ts-ignore
-              '--eds-input-background': '#85babf5e',
-            }
-          : {}
-      }
-    />
+    <>
+      <Typography className={'self-center pb-2'} bold={true}>
+        {label}
+      </Typography>
```
henrikbossart, line 15: Same comment here about the label

packages/dm-core-plugins/src/form/widgets/NumberWidget.tsx
```diff
@@ -1,33 +1,38 @@
...
-        isDirty && props.variant !== 'error'
-          ? {
-              // @ts-ignore
-              '--eds-input-background': '#85babf5e',
-            }
-          : {}
-      }
-    />
+    <>
+      <Typography className={'self-center pb-2'} bold={true}>
```
henrikbossart, line 13: wrap in label?

packages/dm-core-plugins/src/form/utils/getDisplayLabel.tsx
```diff
@@ -4,6 +4,6 @@ export const getDisplayLabel = (attribute: TAttribute): string => {
   const { name, label, optional } = attribute
 
   const displayLabel = label === undefined || label === '' ? name : label
-
-  return optional ? `${displayLabel} (optional)` : displayLabel
+  return displayLabel
+  //return optional ? `${displayLabel} (optional)` : displayLabel;
```
henrikbossart, line 8: Remove?

packages/dm-core-plugins/src/form/templates/ObjectModelUncontainedTemplate.tsx
```diff
@@ -27,53 +27,85 @@ export const ObjectModelUncontainedTemplate = (
...
-        )}
-        {address && onOpen && !uiAttribute?.showInline && (
-          <OpenObjectButton
-            viewId={namePath}
-            viewConfig={{
-              type: 'ReferenceViewConfig',
-              scope: '',
-              recipe: uiRecipe?.name,
+    <div>
+      <legend className='flex flex-row h-10 justify-between border-b-[2px]'>
```
henrikbossart, line 44: `flex-row` is redundant, and the `border-b-2` is already 2px
```suggestion
      <legend className='flex h-10 justify-between border-b-2'>
```

henrikbossart (approved):
All in all this is good, I just have a couple of suggestions 😄

packages/dm-core-plugins/src/form/widgets/NumberWidget.tsx
```diff
@@ -1,33 +1,38 @@
...
-      style={
-        isDirty && props.variant !== 'error'
-          ? {
-              // @ts-ignore
-              '--eds-input-background': '#85babf5e',
-            }
-          : {}
-      }
-    />
+    <>
```
eoaksnes, line 12: We should keep the eds label usage, for consistency. So that we have the same look for all our plugins.

eoaksnes (changes requested):

packages/dm-core-plugins/src/form/widgets/TextareaWidget.tsx
pbullhove: image
Like this?

pbullhove: Is this correct, also setting the id to be the label?

pbullhove: Would not this be more semantically correct, using id instead of label? 
image

packages/dm-core-plugins/src/form/components/SelectReference.tsx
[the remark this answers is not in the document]
pbullhove: That's nice. I like it. 
Is that an EDS icon?

packages/dm-core-plugins/src/form/widgets/TextareaWidget.tsx
henrikbossart: When digging through the dom it looks like EDS applies the label as the id of the textarea. Thats why I suggest using the label in the `htmlFor`. If the id you pass on the `TextField` component gets passed down to the `textfield` tag, you may use that instead.

packages/dm-core-plugins/src/form/components/SelectReference.tsx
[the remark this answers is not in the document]
henrikbossart: Note: this is not an actual icon in the EDS library, but one I made myself using two other icons in the EDS lib.

henrikbossart:
I think I agree with what @eoaksnes is mentioning here about the line makes it look like a header of sorts 🤔

packages/dm-core-plugins/src/form/widgets/NumberWidget.tsx
```diff
@@ -1,33 +1,41 @@
...
+      {/* <label htmlFor={props.id}>
+        <Typography className={"self-center pb-2"} bold={true}>
+          {label}
+        </Typography>
+      </label> */}
```
henrikbossart, line 17:
```suggestion
```

packages/dm-core-plugins/src/form/widgets/TextWidget.tsx
```diff
@@ -9,26 +9,33 @@ const TextWidget = (props: TWidget) => {
...
-            }
-          : {}
-      }
-    />
+    <>
+      {/* <label htmlFor={props.id}>
+        <Typography className={"self-center pb-2"} bold={true}>
+          {label}
+        </Typography>
+      </label> */}
```
henrikbossart, line 17:
```suggestion
```

packages/dm-core-plugins/src/form/widgets/TextareaWidget.tsx
```diff
@@ -13,14 +13,19 @@ const TextareaWidget = (props: TWidget) => {
...
-      rows={5}
-      onChange={onChangeHandler}
-      label={label}
-    />
+    <>
+      {/* <label htmlFor={props.id}>
+        <Typography className={"self-center pb-2"} bold={true}>
+          {label}
+        </Typography>
+      </label> */}
```
henrikbossart, line 21:
```suggestion
```

henrikbossart (approved):
Good, but remove commented code :)

packages/dm-core-plugins/src/form/widgets/TextWidget.tsx
pbullhove: Nice catch

eoaksnes (approved):
👍

PR #925, opened 2023-12-08
feat(edit job): update recipe and add missing form functionality

eoaksnes:

* Pimped Edit Job (this is only the first version)

image

* Refactored `ObjectField` in `FormPlugin` to make it easier to understand
* Added `EntityPickerWidget` to form
* Added `AddObjectBySearchButton` that can create a menu of blueprints to choose from by searching DMSS after a specific blueprint type and finding all blueprints that extend from that type.

image

* Added `readOnly` to `Field` blueprint so that can specifiy indivial if a field (Number, String, Object or Array) should be readonly, but not the whole form

> **NOTE**: Seperate issue on fixing the DateTimeWidget to display read-only in a better way and the EntityPicker needs to be adjusetd.


eoaksnes:
already added this code in another PR

PR #928, opened 2023-12-09
feat(form): open and expand as the same time

eoaksnes:

* Add the possibility of both Expanding and Opening with two different ViewConfig at the same time.

Just specify for a single attribute like this: 

```json
 {
    "name": "something",
    "type": "PLUGINS:dm-core-plugins/form/fields/ObjectField",
    "expandViewConfig": {
      "type": "CORE:ReferenceViewConfig",
      "recipe": "A"
    },
    "openViewConfig": {
      "type": "CORE:ReferenceViewConfig",
      "recipe": "B"
    },
    "showExpanded": true,
    "functionality": {
      "type": "PLUGINS:dm-core-plugins/form/FormFunctionalityConfig",
      "open": true,
      "expand": true
    }
}
```

Adjust default values for all attributes (default is expand=false and open=true):

```json
"config": {
    "type": "PLUGINS:dm-core-plugins/form/FormInput",
    "functionality": {
        "type": "PLUGINS:dm-core-plugins/form/FormFunctionalityConfig",
        "open": true,
        "expand": true
    }
}
```

Example Accountant that can  open and expand at the same time:

image

> **NOTE**: https://github.com/equinor/dm-core-packages/pull/925 needs to be merged inn first.


pbullhove (approved):
Very nice EO! This is great work.

PR #931, opened 2023-12-11
Feat/further form UI

pbullhove:

Since all form objects should have the same look, we refactor the ui into reusable components. 
These are now located under `/shared/`. 

Starting to apply this to the model contained template. 

image


packages/dm-core-plugins/src/form/templates/ObjectModelContainedTemplate.tsx
```diff
@@ -24,24 +24,43 @@ export const ObjectModelContainedTemplate = (
     uiAttribute?.showExpanded ?? config.showExpanded
   )
   const value = watch(namePath)
-  const isCreated = value && Object.keys(value).length > 0
+  const objectIsNotEmpty = value && Object.keys(value).length > 0
   const canExpand =
-    isCreated &&
+    objectIsNotEmpty &&
```
soofstad, line 29: Should this also be true if ever "canOpenInTab" is false. I think so

packages/dm-core-plugins/src/form/templates/shared/FormExpandedVewWrapper.tsx
```diff
@@ -0,0 +1,11 @@
+import React, { PropsWithChildren } from 'react'
+
+const FormExpandedVewWrapper = ({ children }: PropsWithChildren) => {
```
soofstad, line 3: Typo? uwu

soofstad (approved):
Good refactor on consolidating styling!

packages/dm-core-plugins/src/form/templates/shared/FormExpandedVewWrapper.tsx
pbullhove: nice catch

packages/dm-core-plugins/src/form/templates/ObjectModelContainedTemplate.tsx
pbullhove: EO had a pull request fixing this

PR #941, opened 2023-12-11
fix: default tab name on open

### LINKED ISSUE #860: feat(view creator): If the element does not have a name, then show the index.

collinlokken:
Use index and attribute name when opening object without name in new tab

Name is often not a required attribute

Close #860

### DIFF
--- a/packages/dm-core-plugins/src/list/ListPlugin.tsx
+++ b/packages/dm-core-plugins/src/list/ListPlugin.tsx
@@ -108,7 +108,10 @@ export const ListPlugin = (props: IUIPlugin & { config?: TListConfig }) => {
       return
     }
 
-    const view = { ...internalConfig.openViewConfig, label: item?.data?.name }
+    const view = {
+      ...internalConfig.openViewConfig,
+      label: item?.data?.name ?? `${attribute?.name} #${item.index}`,
+    }
     onOpen(
       item.key,
       view,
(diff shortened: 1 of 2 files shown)


packages/dm-core-plugins/src/list/ListPlugin.tsx
```diff
@@ -108,7 +108,12 @@ export const ListPlugin = (props: IUIPlugin & { config?: TListConfig }) => {
       return
     }
 
-    const view = { ...internalConfig.openViewConfig, label: item?.data?.name }
+    const view = {
+      ...internalConfig.openViewConfig,
+      label: item?.data?.name
+        ? item?.data?.name
+        : `${attribute?.name} #${item.index}`,
```
henrikbossart, line 115: I think this can be simplified to this?
```suggestion
      label: item?.data?.name ?? `${attribute?.name} #${item.index}`,
```

packages/dm-core/src/components/Table/TableRow/TableRow.tsx
```diff
@@ -40,7 +40,12 @@ export function TableRow(props: TableRowProps) {
   function openItemAsTab() {
     props.onOpen(
       crypto.randomUUID(),
-      { label: item?.data?.name, type: 'ViewConfig' },
+      {
+        label: item?.data?.name
+          ? item?.data?.name
+          : `${idReference.split('.').slice(-1)} #${item.index}`,
```
henrikbossart, line 46: Same as above
```suggestion
        label: item?.data?.name ?? `${idReference.split('.').slice(-1)} #${item.index}`,
```

packages/dm-core-plugins/src/list/ListPlugin.tsx
collinlokken: Agreed! 👍

eoaksnes (approved):

PR #945, opened 2023-12-12
fix: Creating new config object on every render breaks memo

soofstad:
- Since `Object.is({}, {}) is false.`, the memo wrapped UiPlugins did not work.
- Solved by passing the whole "recipe" object to memoized function, before creating a default {}

- Rerender on hover breaks some child components that can not rerender at any point (plotly etc.)

closes #https://github.com/equinor/dm-app-simpos/issues/65


henrikbossart (approved):
Great! 🎉

PR #946, opened 2023-12-12
fix: dimensional scalar test stabilisation

roar-larsen:
Test is still unstable, trying to add more expects.

If this does not work I suspect same issue as with the model uncontained test (#847)


soofstad:
Have you tried running the CI multiple times? To see if this change had the wanted effect

roar-larsen:
Not this time, but could try running again a few times before merging now.

soofstad (approved):

PR #958, opened 2023-12-13
Use official biome pre-commit hook

henrikbossart:
Replaces the local hook with the official biome pre-commit hook.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -39,10 +39,8 @@ repos:
       - id: conventional-pre-commit
         stages: [ commit-msg ]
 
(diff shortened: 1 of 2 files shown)


eoaksnes (approved):

soofstad (changes requested):
This is still a "local" repo hook :o 
Can have a look at the link provided in the issue on how it should be added.

henrikbossart:
I'm really struggling to make this file ignoring work. Could someone with more pre-commit experience please take a look at my pattern? 🙏

soofstad (approved):

PR #960, opened 2023-12-13
fix: update nested form

eoaksnes:

* Fix the bug that nested forms that are inline (expanded) do not update themselves.

The fix was to include them in the payload that was sent to the API.


henrikbossart (approved):

pbullhove (approved):
Good.

PR #970, opened 2023-12-13
feat: readonly primitives new ui looks better

pbullhove:

Readonly now looks more like an inputField. 
Is a custom component. 
image


eoaksnes (changes requested):
The readonly field should be a widget.

eoaksnes (approved):

eoaksnes:
Maybe just keep it, it is more flexible to have it inside the widgets. 

The alternative is to do something like this inside the  StringField and Numberfield

```
const Widget = getWidget(uiAttribute?.widget ?? (readOnly ? 'ReadOnlyWidget : ' TextWidget'))
```

PR #971, opened 2023-12-13
refactor: reuse colors

pbullhove:

Putting colors in tailwind config and using them with tailwind instead of hex codes directly

Also added a colors.ts that can be imported and used in styled components like 
```
import {colors} from '@development-framework/dm-core'
const Cool = styled(div)`
  color: ${colors.background};
`
```

❓ Was not able to direcly import these into tailwind config in 'example' and 'dm-core-plugins'. Don't understand why.


collinlokken (approved):
Nice 😄

PR #977, opened 2023-12-14
fix: job label and close datepicker on click outside

soofstad:
- DatePicker now closes when you click outside it
- If the jobConfig has a label, use it when rendering


collinlokken (approved):
LGTM 👍

PR #978, opened 2023-12-14
Feat/inline dim skalar

pbullhove:

image

Everything about it is configurable. 

Further refactoring making this inline - primitive into a reusable component will come later.


packages/dm-core-plugins/src/form/widgets/TextWidget.tsx
```diff
@@ -26,7 +26,7 @@ const TextWidget = (props: TWidget) => {
               // @ts-ignore
               '--eds-input-background': '#85babf5e',
             }
-          : {}
+          : { borderRadius: '10px !important' }
```
collinlokken, line 29: 🥶

packages/dm-core-plugins/src/form/widgets/DimensionalScalarWidget.tsx
```diff
@@ -1,8 +1,19 @@
...
+  EdsProvider,
+  TextField,
+  Tooltip,
+  Typography,
+} from '@equinor/eds-core-react'
 import { NumberFieldWithoutArrows } from '../components/NumberFieldWithoutArrows'
 
+const parseWidthString = (widthString: string) => {
+  const width = parseFloat(widthString)
+  if (!width || width < 0) return
```
collinlokken, line 13: Can width be negative?

packages/dm-core-plugins/src/form/widgets/TextWidget.tsx
pbullhove: lol

PR #980, opened 2023-12-14
feat: bugFix + disabled look + refactor: input fields 

pbullhove:

Feat: 
- Readonly looks better, like specified in earlier (but not merged) PR

FIX: 
- scrolling on a numberfield does not increment the numbers, as it caused increasing when just scrolling down the page. 

Refactor: 
- move some of the reused code in calling the numberField and textfield into reusable components. 
- now only StyledNumberField and StyledTextField are exported, both inheriting from a general EDSField.


packages/dm-core-plugins/src/form/components/StyledInputFields.tsx
```diff
@@ -0,0 +1,80 @@
...
+  )
+}
+
+export const StyledTextField = (
+  props: React.ComponentProps<typeof StyledEDSField>
+) => {
+  return (
+    <StyledEDSField
+      {...props}
+      data-testid={`form-text-widget-${props.label}`}
```
eoaksnes, line 48: data-testid=form-text-widget-{...}. I guess this code should be inside the TextWidget or be renamed to form-text-input-{...}? I guess the question is, do we need to have a dedicated data-testid pr widget or not?

packages/dm-core-plugins/src/form/components/StyledInputFields.tsx
```diff
@@ -0,0 +1,80 @@
...
+  props: React.ComponentProps<typeof StyledEDSField>
+) => {
+  return (
+    <NumberFieldWithoutArrows
+      {...props}
+      onWheel={(event: React.UIEvent<HTMLInputElement>) =>
+        (event.target as HTMLInputElement).blur()
+      }
+      type={'number'}
+      data-testid={`form-number-widget-${props.label}`}
```
eoaksnes, line 77: same as above

eoaksnes (approved):
👍

eoaksnes (approved):

PR #987, opened 2023-12-15
feat: multiple templates

eoaksnes:

* Add the possibility to have multiple templates to select from when creating new items in a list or table plugin.

Table:

image

image

List:

image

> **NOTE**: If only a single template is provided the menu is not shown, and the item is automatically picked as a template.


pbullhove (approved):
Didn't look too thoughly

PR #996, opened 2023-12-18
feat: Form reset button + further ui changes

pbullhove:
## Added form reset button that resets all dirty states

image

## Complex Objects Ui change

1. Border is lighter
2. Border is slightly more rounded. 
image

## Primitive Array Ui change

1. No padding around it 
2. The container spans the same width as the array now. 
3. Made background white 

image

## Form Changes

1. More padStart in the input box. 
2. Less padding above the form and between the elements
4. Submit button is diabled when there are no changes to save. 
image

image


packages/dm-core-plugins/src/form/widgets/DimensionalScalarWidget.tsx
```diff
@@ -55,11 +55,11 @@ const DimensionalScalarWidget = (props: TWidget) => {
         style={
           widgetConfig?.inline
             ? {
-                border: 'solid #6c6c6c 1px',
+                // borderBottom: 'solid #6c6c6c 1px',
                 borderRadius: '2px 0 0 2px',
                 display: 'flex',
                 alignItems: 'center',
-                background: '#dedede',
+                // background: '#f7f7f7',
```
soofstad, line 62: Could we group these colors somewhere?
Many hard coded hex codes...

soofstad (approved):

packages/dm-core-plugins/src/form/widgets/DimensionalScalarWidget.tsx
pbullhove: yes, there is some colors in tailwind.config.

pbullhove: now uses reusable colors bg-equinor-lightgray

PR #1001, opened 2023-12-18
Fix(datepicker): input bugs

henrikbossart:
- Fixes bug that showed the time section in the input field even though 'date' variant was used.
- Removes line underneath calendar if only date is used
- Sets "overshot" day/month to closest valid day/month.


packages/dm-core/src/components/Pickers/Datepicker/Datepicker.tsx
```diff
@@ -199,13 +201,16 @@ export const Datepicker = (props: DatepickerProps): ReactElement => {
           }`}
           style={{ zIndex: 9999, width: '25rem' }}
         >
+          <button type='button' className='bg-equinor-lightgreen'>
+            button
+          </button>
```
soofstad, line 206: What does this do? :thinking:

soofstad (approved):

packages/dm-core/src/components/Pickers/Datepicker/Datepicker.tsx
henrikbossart: Click it, and you will see 🤓

PR #1013, opened 2023-12-18
fix: set dirty state after update

eoaksnes:

* fix: set the state as dirty on update so that save button is enabled


l3abak (approved):
ok

PR #1014, opened 2023-12-18
fix: multiple templates not save when add

eoaksnes:

* fix: save added row to DMSS when adding a item in not edit mode


l3abak (approved):
ok

PR #1017, opened 2023-12-19
fix: add indexLabel to list and table plugin

### LINKED ISSUE #965: feat(table, list): has possibility to show index number by config

collinlokken:
Closes #965 

image (1)
image (2)
image

### DIFF
--- a/packages/dm-core/src/components/Table/TableRow/TableRow.tsx
+++ b/packages/dm-core/src/components/Table/TableRow/TableRow.tsx
@@ -14,6 +14,7 @@ import * as Styled from './styles'
 import { TableRowActions } from './TableRowActions/TableRowActions'
 import { TableCell } from './TableCell/TableCell'
 import { add } from '@equinor/eds-icons'
+import { toast } from 'react-toastify'
 
 export function TableRow(props: TableRowProps) {
   const {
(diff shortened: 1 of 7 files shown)


roar-larsen:
Should the index start at 1 instead of 0?

image

packages/dm-core/src/components/Table/TableRow/TableRow.tsx
```diff
@@ -47,12 +48,19 @@ export function TableRow(props: TableRowProps) {
   }
 
   function openItemAsTab() {
+    if (!props.onOpen) {
+      toast.error(
+        'Invalid UiRecipes. The table plugin was not passed an "onOpen()"-function.'
+      )
+      return
+    }
```
soofstad, line 56: Should be allowed not to pass an "onOpen" function

soofstad (changes requested):
Not a big fan of this "indexLabel" parameter. 
I would rather see defaults that works good enough.
But if this was decided with PO I guess it's fine

packages/dm-core/src/components/Table/TableRow/TableRow.tsx
collinlokken: testet bare kjapt med `if(true) ...` på linje 51
image

soofstad:
Would like your opinion here @l3abak 

In my opinion, this "captionByIndex" and "caption" are misplaced in the uiRecipe for ListPlugin. If they belong there, they belong in any plugin, since any plugin can be passed the "onOpen()"-function.

If they truly are plugin config parameters needed everywhere, I think we should find another solution.

And even better, we can generate these captions with some default rules, that will make for a good enough solution for 90% of the cases.
  - For lists; attribute name of list + index
  - For objects: entity.label ?? entity.name ?? attribute name

soofstad (approved):

PR #1022, opened 2023-12-19
fix: allow list header to grow with flexbox

### LINKED ISSUE #991: List plugin overflow UI bug
when name / headers becomes too long then the buttons are pushed to the right. 

Image

collinlokken:
- Allow ListPlugin to shrink down at the minimum as wide as its content
- Use flexbox on list items (rows)

Closes #991 

image
image
image


soofstad (approved):
:+1:

PR #1023, opened 2023-12-19
fix: use field name in payload

awesthouse:
Uses correct fieldName in payload, not hard-coded field "data"

Trying to post to data field, when it might not exist


eoaksnes (approved):
👍

PR #1027, opened 2023-12-19
fix: compact list

pbullhove:
Made a flag to control list compactness. 
Either list is compact or "normal" 

Also made the overflow make sense, with overflow ellipsis if the list attributes span too far, and no overflow. 

image

Compared to non-compact: 
image


collinlokken (approved):
LGTM 👍  A suggestion is to maybe have scroll bar when compact is not used?

PR #1032, opened 2023-12-20
style: fix many minor style issues

soofstad:
- Toggling "job is recurring" does not hide the toggle component
- Advanced schedule view is hidden by default
- Setting advanced schedule hides "simple" input, and vice versa
- Make the log itself scroll able on overflow, not the whole modal hiding buttons and title
- When media is loading, don't show message "Not found", but "Loading..."
- Color background of selected menuItem (Sidebar)
- Center copy buttons in Yaml

### DIFF
--- a/packages/dm-core-plugins/src/yaml/YamlPlugin.tsx
+++ b/packages/dm-core-plugins/src/yaml/YamlPlugin.tsx
@@ -58,7 +58,7 @@ const YamlView = (props: {
 
   return (
     <div>
-      <div className='flex justify-end items-end my-2 gap-1'>
+      <div className='flex justify-end items-center my-2 gap-1'>
         <Button variant='ghost' onClick={() => onClick(asYAML)}>
           Copy as YAML
         </Button>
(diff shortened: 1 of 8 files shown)


packages/dm-core-plugins/src/yaml/YamlPlugin.tsx
```diff
@@ -58,7 +58,7 @@ const YamlView = (props: {
 
   return (
     <div>
-      <div className='flex justify-end items-end my-2 gap-1'>
+      <div className='flex justify-end items-center my-2 gap-1'>
```
henrikbossart, line 61: Doesn't this make it unaligned with the depth input field? 🤔 I seem to recall the reason I aligned it to the end was because the label on the input field made the centering not look nice.

henrikbossart (approved):
👍

packages/dm-core-plugins/src/yaml/YamlPlugin.tsx
soofstad: It was unaligned. Now its a bit better, still not perfect :/ 
The buttons did just change, your alignment might have been before that?

henrikbossart: aaah, probably

PR #1044, opened 2023-12-21
fix: error when selecting templates for uncontained

pbullhove:

FIX templates for list
- Makes it so uncontained list does not try to use templates, which caused errors. 
- Also makes contained use template[0] if there is only one template provided. 

REFACTOR templates for table

- code here was a bit messy, so refactored it to same pattern as used for lsit


soofstad (approved):
Looks good!

PR #1048, opened 2023-12-21
Feat: better list padding + default pagination view

pbullhove:

Config to set default pagination. 
- is put into sorted list of paginationOptions
- if numItems < Min(defaultPaginationSize, selectedPaginationSize) then don't show pagination selector. 

Video of how it works: 
https://jam.dev/c/885b812e-70e2-4a17-b4a7-5a84042f13f5

### DIFF
--- a/packages/dm-core-plugins/blueprints/list/ListPluginConfig.json
+++ b/packages/dm-core-plugins/blueprints/list/ListPluginConfig.json
@@ -97,6 +97,13 @@
       "description": "Override label for all items that are opened in a ViewSelector from List.",
       "attributeType": "string",
       "optional": true
+    },
+    {
+      "name": "defaultPaginationRowsPerPage",
+      "type": "CORE:BlueprintAttribute",
+      "description": "The number of rows per page to show by default. If there are less rows in the list than this number, then pagination selector will not be visible. ",
+      "attributeType": "number",
+      "optional": true
     }
   ]
 }
--- a/packages/dm-core-plugins/src/list/ListPlugin.tsx
+++ b/packages/dm-core-plugins/src/list/ListPlugin.tsx
@@ -47,6 +47,7 @@ type TListConfig = {
   resolveReferences: boolean
   templates?: TTemplate[]
   labelByIndex?: boolean
+  defaultPaginationRowsPerPage?: number
   label?: string
 }
 const defaultConfig: TListConfig = {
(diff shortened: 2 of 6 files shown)


packages/dm-core-plugins/blueprints/list/ListPluginConfig.json
```diff
@@ -97,6 +97,13 @@
...
       "attributeType": "string",
       "optional": true
+    },
+    {
+      "name": "defaultPaginationRowsPerPage",
+      "type": "CORE:BlueprintAttribute",
+      "description": "The number of rows per page to show by default. If there are less rows in the list than this number, then pagination selector will not be visible. ",
+      "attributeType": "number",
+      "optional": true
     }
```
awesthouse, line 107: Maybe this should show the default (5) if it is actually set in the ListPlugin?

packages/dm-core-plugins/src/list/ListPlugin.tsx
```diff
@@ -59,7 +59,7 @@ const defaultConfig: TListConfig = {
   selectFromScope: undefined,
   hideInvalidTypes: false,
   compact: false,
-  defaultPaginationRowsPerPage: 10,
+  defaultPaginationRowsPerPage: 5,
```
awesthouse, line 62: Was it requested that we universally change this to a smaller amount? 🙂 Just checking cause this will change the default on anything that uses List

packages/dm-core-plugins/src/list/ListPlugin.tsx
```diff
@@ -93,17 +93,24 @@ export const ListPlugin = (props: IUIPlugin & { config?: TListConfig }) => {
...
 
-  const defaultPaginationRowsPerPage =
-    internalConfig.defaultPaginationRowsPerPage ?? 10
-  const showPagination = useMemo(
-    () => items.length > defaultPaginationRowsPerPage,
-    [items]
-  )
+  const defaultPaginationRowsPerPage = useMemo(() => {
+    let numRows = internalConfig.defaultPaginationRowsPerPage ?? 5
+    numRows = Math.round(numRows)
```
awesthouse, line 98: I know we joked about it in the meeting, but is it maybe not better to expect user to input a default that makes sense than do 2 operations like this when it might not be necessary? 😅

awesthouse (approved):
Nice! A lot of good changes 🧠💪🏻

PR #1049, opened 2023-12-21
fix: default items per page should be 10

pbullhove:

Babak said default number of items should be 10


awesthouse (approved):
👍🏻

PR #1054, opened 2023-12-21
fix: table opens too many tabs

### LINKED ISSUE #1055: fix(table): table opens too many tabs

collinlokken:
Use a deterministic key to open tabs
image

Table duplicates tabs because a different viewId is used for each open event

Closes #1055


soofstad (approved):
Good catch!

henrikbossart (approved):

PR #1062, opened 2024-01-03
test: added test for table templates

roar-larsen:
Added test using table templates.


eoaksnes (approved):
👍

PR #1067, opened 2024-01-03
fix: update existing viewitem on add

### LINKED ISSUE #1065: fix(table): reorder the rows does not update the index used for tabs
Reordering the table does not change the index used for tabs, deleting or reordering.

Image

collinlokken:
Updates the tab (viewItem) with the latest state onOpen (e.g. with new index when position has changed after delete)

Opening tab after delete does not update the index

Closes #1065

### DIFF
--- a/packages/dm-core-plugins/src/view_selector/useViewSelector.tsx
+++ b/packages/dm-core-plugins/src/view_selector/useViewSelector.tsx
@@ -52,19 +52,27 @@ export function useViewSelector(
     onSubmitAdded?: (data: any) => void,
     onChangeAdded?: (data: any) => void
   ) => {
-    if (!viewSelectorItems.find((view: TItemData) => view.viewId === viewId)) {
+    const existingViewIndex = viewSelectorItems.findIndex(
+      (view: TItemData) => view.viewId === viewId
+    )
+    const newView: TItemData = {
+      viewId: viewId,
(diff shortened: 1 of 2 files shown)


packages/dm-core-plugins/src/view_selector/useViewSelector.tsx
```diff
@@ -52,19 +52,25 @@ export function useViewSelector(
...
-        rootEntityId: rootId || idReference,
-        onSubmit: onSubmitAdded,
-        onChange: onChangeAdded,
-        closeable: true,
-        isSubItem: isSubItem,
-      }
       setViewSelectorItems([...viewSelectorItems, newView])
+    } else {
+      newView.viewId = viewSelectorItems[existingViewIndex].viewId
+      viewSelectorItems[existingViewIndex] = newView
```
soofstad, line 73: How can this work, when you are not calling "setViewSelectorItems()"?

soofstad (changes requested):

packages/dm-core-plugins/src/view_selector/useViewSelector.tsx
collinlokken: It works, but mutating state is not recommended. Setting state again is better 👍

soofstad (approved):

PR #1068, opened 2024-01-03
fix: select entity label as view label

### LINKED ISSUE #1064: fix(table, list): correct order of defaults when indexing
The naming does not use the correct order of defaults when indexing:
If indexing is enabled, i.e. labelByIndex true, the order of label should be 1) label inside Viewconfig, 2)label inside the entity, 3)name inside the entity, 4)list attribute name

collinlokken:
Prioritieses entitty label above enity name when generating label for viewSelectorItem

Requested feature

Closes #1064

### DIFF
--- a/packages/dm-core-plugins/src/list/ListPlugin.tsx
+++ b/packages/dm-core-plugins/src/list/ListPlugin.tsx
@@ -150,11 +150,11 @@ export const ListPlugin = (props: IUIPlugin & { config?: TListConfig }) => {
       )
       return
     }
-    const label = config.label
-      ? config.label
-      : item?.data?.name
-        ? item?.data?.name
-        : `${attribute?.name}`
+    const label =
+      config.label ||
+      item.data?.label ||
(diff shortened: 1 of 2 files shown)


packages/dm-core-plugins/src/list/ListPlugin.tsx
```diff
@@ -152,9 +152,11 @@ export const ListPlugin = (props: IUIPlugin & { config?: TListConfig }) => {
...
     const label = config.label
       ? config.label
-      : item?.data?.name
-        ? item?.data?.name
-        : `${attribute?.name}`
+      : item?.data?.label
+        ? item?.data?.label
+        : item?.data?.name
+          ? item?.data?.name
+          : `${attribute?.name}`
```
soofstad, line 159: Shorten with "||" instead of " ? :"?

soofstad (changes requested):

soofstad (approved):

PR #1069, opened 2024-01-03
Fix/media plugin bugs

henrikbossart:
- Uses `<embed>` `<iframe>` for showing files, to prevent downloading it.
- Adds the date back into the meta
- Adds a disclaimer for unknown filetypes

Screenshot 2024-01-04 at 09 41 19
Screenshot 2024-01-04 at 09 41 26


packages/dm-core/src/components/Pickers/Datepicker/Datepicker.tsx
```diff
@@ -77,6 +77,7 @@ export const Datepicker = (props: DatepickerProps): ReactElement => {
     if (dateInput) {
       const { day, month, year, max } = extractDateComponents(dateInput)
       const convertedDate = DateTime.utc(year, month, day)
+      console.log(convertedDate)
```
soofstad, line 80:
```suggestion
```

packages/dm-core/src/utils/filetypes.ts
```diff
@@ -0,0 +1,20 @@
...
+  'mp4',
+  'mov',
+  'wmv',
+  'webm',
+  'avi',
+  'flv',
+  'mkv',
+  'mts',
+]
+export const applicationFiletypes = ['pdf']
```
soofstad, line 20: Is "application" a bad name? :thinking: 
How about, 'pdf'? Or are there other examples?

packages/dm-core/src/components/MediaContent.tsx
```diff
@@ -69,24 +76,44 @@ export const MediaContent = (props: MediaContentProps): ReactElement => {
...
+            Unknown filetype
+          </h2>
+          <p className='flex gap-1 mb-3'>
+            The filetype <code>{meta.filetype}</code> is not known, and can not
+            be previewed. You may still download the file
+          </p>
+          <Button download={`${meta.title}.${meta.filetype}`} href={blobUrl}>
+            <Icon size={16} data={download} />
+            Download
+          </Button>
```
soofstad, line 107: I would change the wording here a bit.
The filetype is not "unknown", we just don't have a compatible viewer for it.
So something in the lines of;
"No preview available. Download the file to open it with another application."

example/app/data/DemoDataSource/plugins/media_viewer/assets/random.bin
soofstad, line 1: Do stuff work if "random.bin" did not have a suffix? As a regular binary file would

soofstad (changes requested):

packages/dm-core/src/utils/filetypes.ts
henrikbossart: right, I initially thought I could include things like "doc" and "docx" here as well, but that didn't work. I'll change the name

example/app/data/DemoDataSource/plugins/media_viewer/assets/random.bin
henrikbossart: It would just show the message but with a blank filetype, I've changed it to this:

Screenshot 2024-01-04 at 14 06 50

soofstad: Nice :+1: . "cannot shown"....

henrikbossart: heh........ missing a 🐝  there

packages/dm-core/src/components/MediaContent.tsx
```diff
@@ -69,24 +72,48 @@ export const MediaContent = (props: MediaContentProps): ReactElement => {
...
+        >
+          <h2 className='text-lg text-equinor-green font-medium'>
+            No preview available
+          </h2>
+          <p className='flex gap-1 mb-3 items-center'>
+            A preview for{' '}
+            <code className='text-sm'>
+              {meta.filetype.length > 0 ? meta.filetype : 'binary'}
+            </code>{' '}
+            files cannot shown. Please download the file and open it in the
```
henrikbossart, line 101:
```suggestion
            files cannot be shown. Please download the file and open it in the
```

soofstad (approved):

henrikbossart:
tests fail because a package used is an ES module. Haven't succeeded in making it work with jest. Am grateful for help, if not I'll look for a different package to hopefully achieve the same.

henrikbossart:
I've been looking into this issue for a while now, but I'm not really getting anywhere.. I think the culprit is the `mime` package, which I use to determine the content-type of the media based off the filetype.

the package is a pure ESModule package, which doesn't really matter, but the jest framework complains about this, and I couldn't get their workarounds to work.

I thought about switching the entire `dm-core` package over to ESM, but that turned out to be an insane amount of work, so yeah... 🤷‍♂️

PR #1077, opened 2024-01-04
Fix/pagination

pbullhove:

1. Refactor usePagination into reusable component
2. When adding new item to list, it moves to last page so that one can see the new item. 
3. Fixed bug where pagination was messed up when on 2. page and changing itemsPerPage

will use this hook in TablePlugin as well, but that is next PR.


henrikbossart:
Should it show which page you are currently on? 🤔 

Screenshot 2024-01-04 at 16 37 22

henrikbossart (approved):
Nice 🚀

pbullhove:
> Should it show which page you are currently on? 🤔
> 
> Screenshot 2024-01-04 at 16 37 22

Perhaps?

PR #1078, opened 2024-01-05
Fix: 2 table pagination bugs

pbullhove:

1. Uses the new hook "usePagination" in table as well, that fixes an old pagination bug
- The bug was that being on last page and chaning items per page made pagination overflow. 

2. Fix sorting bug: 
- Makes table jump to page 0 when sorting. Earlier sorting put the first elements on the top of the current page, not the top of the whole table, which was a bug.


soofstad (approved):
Good stuff!

PR #1088, opened 2024-01-08
fix: use form state to update entity

### LINKED ISSUE #1073: bug(primitive-array): some indexing error.
See video: 
https://jam.dev/c/18c2a7a4-1a82-483a-997c-d84ea14237be

Steps to reproduce:
- Add two new rows
- Populate one with data
- Submit and get required-error
- Remove the one with error
- Submit again.

collinlokken:
Use react hook from context to manipulate form state when deleting item in primitive array

Bug where primitive array is not updated correctly when deleting while in dirty state

Closes #1073

### DIFF
--- a/packages/dm-core-plugins/src/form/components/PrimitiveArray.tsx
+++ b/packages/dm-core-plugins/src/form/components/PrimitiveArray.tsx
@@ -7,6 +7,7 @@ import {
   DeleteSoftButton,
   TAttribute,
 } from '@development-framework/dm-core'
+import { useFormContext } from 'react-hook-form'
 
 interface PrimitiveArrayProps {
   uiAttribute: TUiAttributeObject | undefined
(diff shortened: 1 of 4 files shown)


packages/dm-core-plugins/src/form/components/PrimitiveArray.tsx
```diff
@@ -39,7 +39,7 @@ const PrimitiveArray = ({
   const { getValues } = useFormContext()
 
   const updateValues = (index: number, newValue: TPrimitive): void => {
-    const newValues = [...getValues(namePath)]
+    const newValues = getValues(namePath) ? getValues(namePath) : []
```
awesthouse, line 42: Pretty sure this should still be a copy, no? We're still modifying the array directly.
Also, shorthand `getValues(namepath) || []`? 🙂

packages/dm-core-plugins/src/form/components/PrimitiveArray.tsx
```diff
@@ -55,7 +55,7 @@ const PrimitiveArray = ({
     }
   }
   const removeItem = (index: number) => {
-    const newValues = [...getValues(namePath)]
+    const newValues = getValues(namePath) ? getValues(namePath) : []
```
awesthouse, line 58: Same as comment in updateValues function

packages/dm-core-plugins/src/form/components/PrimitiveArray.tsx
collinlokken: Will apply the shorthand 👍 
Modifying the return value of getValues does not cause a re-render, so it is not necessary to perform a deep copy.

awesthouse (approved):
🎉

PR #1101, opened 2024-01-09
feat: name in new entity dialoig

### LINKED ISSUE #1063: fix: be able to specify name when creating a entity in explorer if the blueprint has a name attribute
1. Add a input field to specify a name for the entity

Image

collinlokken:
Adds name text input to new entity dialogue if blueprint has name attribute
image
image
image

- Fall back to "new_entity" as name if name field is empty when creating

Closes #1063

### DIFF
--- a/packages/dm-core-plugins/src/explorer/components/dialogs/NewEntityDialog.tsx
+++ b/packages/dm-core-plugins/src/explorer/components/dialogs/NewEntityDialog.tsx
@@ -1,12 +1,16 @@
 import {
+  ApplicationContext,
   BlueprintPicker,
   Dialog,
   ErrorResponse,
+  TAttribute,
+  TBlueprint,
   TreeNode,
+  useDMSS,
 } from '@development-framework/dm-core'
-import { Button, Progress } from '@equinor/eds-core-react'
+import { Button, Progress, TextField } from '@equinor/eds-core-react'
 import { AxiosError } from 'axios'
-import React, { useState } from 'react'
+import React, { useContext, useEffect, useState } from 'react'
 import { toast } from 'react-toastify'
 import { EDialog } from '../../types'
 import {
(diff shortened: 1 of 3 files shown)


packages/dm-core-plugins/src/explorer/components/dialogs/NewEntityDialog.tsx
```diff
@@ -14,6 +16,7 @@ import {
   STANDARD_DIALOG_WIDTH,
 } from '../context-menu/NodeRightClickMenu'
 
+const defaultName = 'new_entity'
```
soofstad, line 19: Could add a random suffix to avoid conflicts

packages/dm-core-plugins/src/explorer/components/dialogs/NewEntityDialog.tsx
```diff
@@ -22,13 +25,15 @@ type TProps = {
 
 const NewEntityDialog = (props: TProps) => {
   const { setDialogId, node, setNodeOpen } = props
-  const [blueprint, setBlueprint] = useState<string>('')
+  const [blueprintName, setBlueprintName] = useState<string>('')
+  const [newName, setNewName] = useState<string>(defaultName)
   const [loading, setLoading] = useState<boolean>(false)
+  const { blueprint } = useBlueprint(blueprintName)
```
soofstad, line 31: This will lead to some 404 before a blueprint is selected? Move it to a useEffect or something?

packages/dm-core-plugins/src/explorer/components/dialogs/NewEntityDialog.tsx
```diff
@@ -57,13 +62,24 @@ const NewEntityDialog = (props: TProps) => {
...
+          formData={blueprintName}
           fieldType={'input-field'}
         />
+        {blueprint?.attributes.filter(
+          (attr: TAttribute) => attr.name === 'name'
+        ) && (
+          <TextField
+            id={'entity-name'}
+            onChange={(e: any) => setNewName(e.target.value)}
+            label={'New name'}
```
soofstad, line 75:
```suggestion
            label={'Name'}
```

packages/dm-core-plugins/src/explorer/components/dialogs/NewEntityDialog.tsx
```diff
@@ -57,13 +62,24 @@ const NewEntityDialog = (props: TProps) => {
...
+            helperText={'Optional'}
```
soofstad, line 76: How do you know it's optional? Should check the blueprint attribute

soofstad: Or just remove it since you set a default name anyway

soofstad (changes requested):

packages/dm-core-plugins/src/explorer/components/dialogs/NewEntityDialog.tsx
collinlokken: removed it completely

packages/dm-core-plugins/src/explorer/components/dialogs/NewEntityDialog.tsx
collinlokken: Added a check to see if it is required 👍

soofstad (approved):

PR #1114, opened 2024-01-10
fix(list): check for scope before trying to resolve potentially undefined

soofstad:
- check for scope before trying to resolve potentially undefined

bug

fixes this: 
image


eoaksnes (approved):

packages/dm-core-plugins/src/list/ListPlugin.tsx
```diff
@@ -74,6 +74,8 @@ const defaultConfig: TListConfig = {
 
 export const ListPlugin = (props: IUIPlugin & { config?: TListConfig }) => {
   const { idReference, config, type, onOpen } = props
+  if (!idReference)
+    throw new Error('idReference is missing when calling ListPlugin!')
```
soofstad, line 78:
```suggestion
```

PR #1115, opened 2024-01-10
fix: hide optional

pbullhove:

Attributes can hide optional with config


packages/dm-core-plugins/src/form/fields/StringField.tsx
```diff
@@ -3,13 +3,14 @@ import { Controller } from 'react-hook-form'
...
-import { getDisplayLabelWithOptional } from '../utils/getDisplayLabel'
+import { getDisplayLabel } from '../utils/getDisplayLabel'
 
 export const StringField = (props: TField) => {
-  const { namePath, uiAttribute, attribute, backgroundColor } = props
+  const { namePath, uiAttribute, attribute } = props
   const Widget = getWidget(uiAttribute?.widget ?? 'TextWidget')
   const { config } = useRegistryContext()
   const readOnly = uiAttribute?.readOnly || config.readOnly
+  console.log(uiAttribute)
```
awesthouse, line 13: 😎

packages/dm-core-plugins/src/form/templates/ArrayPrimitiveDataGridTemplate.tsx
```diff
@@ -43,7 +43,7 @@ export const ArrayPrimitiveDatagridTemplate = (
   return (
     <Fieldset>
       <Legend data-testid='primitive-array-legend'>
-        <Typography bold={true}>{getDisplayLabel(attribute)}</Typography>
+        <Typography bold={true}>{getDisplayLabel(attribute, true)}</Typography>
```
awesthouse, line 46: May I ask why so many fields are set to default hide the optional label? I'm sure there's a reason, just a bit confusing 🙂

awesthouse (approved):
Still feel like it's better and more common practice to mark required fields with * , but this looks good for the solution we have now. I don't really understand what the argument of doing this could be in terms of like user experience? Would love.a short explanation if you have time 👍🏻

packages/dm-core-plugins/src/form/templates/ArrayPrimitiveDataGridTemplate.tsx
awesthouse: I say many, but it's not a lot! I just feel like these CAN be optional, no?

packages/dm-core-plugins/src/form/fields/StringField.tsx
pbullhove: Thanks!

packages/dm-core-plugins/src/form/templates/ArrayPrimitiveDataGridTemplate.tsx
pbullhove: Hmm I think you cannot delete completely a primitive array? 
You simply delete all elements? Perhaps I am mistaken.

pbullhove:
> Still feel like it's better and more common practice to mark required fields with * , but this looks good for the solution we have now. I don't really understand what the argument of doing this could be in terms of like user experience? Would love.a short explanation if you have time 👍🏻

Well.. I don't have string opinions. 
I think the (Optional) thing could easily be replaced by a * for the required ones. 
Also it's starting to be a bit messy with the form trying to "know" things about the data and appending optional, since babak wants to be able to turn it off as well. 

Also like, can a checkbox be optional? What does that mean, does it mean that it must be checked?
So many of these were never shown. 


Also for the complex types  the optional is not in the label direcly, but appended after as a badge for ui purposes...

So it's not the best. We should have a short session and figure out what do to about this. 
Because moving forward, everything should be configurable.

packages/dm-core-plugins/src/form/templates/ArrayPrimitiveDataGridTemplate.tsx
awesthouse: Huuuuh, I don't know! 😅 You probably should be able to remove it if it is optional (even tough it doesn't look like you can delete after instantiating right now)
Here the Numbers only field is optional, it should probably say so? I mean, if we're marking things that are optional with optional, we should probably tag all things that are optional with optional otherwise I feel like it's inconsistent and confusing for the user

Skjermbilde 2024-01-10 kl  14 45 07

awesthouse:
Definitely complicated! Yes, required for checkboxes in forms usually means that user HAS to "accept" or click it to be true. For an attribute required it means that we expect a value, whether it's true or false, but in form required for checkboxes actually mean that you expect it to be true. Should probably have a short session, yes 🙂 But I still think this is good for what was requested

packages/dm-core-plugins/src/form/templates/ArrayPrimitiveDataGridTemplate.tsx
pbullhove: Fair point, I'll change it.

Issue #1116, opened 2024-01-10
test: create entity with required name


collinlokken:
Test if the "create" button is disabled when trying to create an entity using a blueprint with a required name field

![Image](https://github.com/equinor/dm-core-packages/assets/43639886/4dab1880-1ce6-4c14-bc4b-8f9ef569525d)

PR #1118, opened 2024-01-11
test(explorer): create entity with required name

roar-larsen:
Expanded test to include checking if the "create" button is disabled when trying to create an entity using a blueprint with a required name field.

#1116


collinlokken (approved):
Nice!

PR #1120, opened 2024-01-11
fix: disable sort when first or last, enable add on 1d primitive

awesthouse:
- Enable add button when data is 1D primitive and dimensions are not defined
- Disable sort up/down respectively when selected item is first or last of list

- Add should be enabled on 1D primitives with undefined dimensions
- Sorting allows you to move row to an index that doesn't really exist


pbullhove (approved):
💰💰💰💰💰 LGTM!!  💰💰💰💰💰

PR #1123, opened 2024-01-11
feat: can hide delete now

pbullhove:

1. form config "hideDelete" makes entity un-deletable even if it is optional 
2. Currently only for storage Contained

image


packages/dm-core-plugins/src/form/templates/ObjectModelContainedTemplate.tsx
```diff
@@ -33,6 +33,7 @@ export const ObjectModelContainedTemplate = (
     uiAttribute,
     onOpen
   )
+  console.log(uiAttribute)
```
soofstad, line 36:
```suggestion
```

soofstad (approved):
:+1:

PR #1125, opened 2024-01-12
feat(job): create job plugin supports templates

### LINKED ISSUE #1102: refactor(job): refresh 'create-from-recipe'
- [ ] Should create jobs from templates instead of this custom uiRecipe
- [ ] Rename to "job/create-from-template"
- [ ] See if we can merge the job plugins to one, or at least share some more code (Like the control button)

soofstad:
-renamed plugin `job/single_job` -> `job/create`
- renamed plugin `job/controls` -> `job`
- plugin "job/create" now uses templates to create new jobs
- both plugins got option to hideLogs
- More shared components and logic between the two plugins
- the "JobButton" is now orange when loading
- "resolveRelativeAddressSimple" now suports ".attribute", ".", and "~.~" like formats

closes #1102


packages/dm-core-plugins/src/job/JobCreate.tsx
```diff
@@ -0,0 +1,213 @@
...
+    if (!jobDocument) return
+    if (asCronJob || jobDocument.type === EBlueprint.RECURRING_JOB)
+      // @ts-ignore
+      setSchedule(jobDocument.schedule?.cron ? jobDocument?.schedule : schedule)
+    if (jobDocument.type === EBlueprint.RECURRING_JOB) setAsCronJob(true)
+  }, [jobDocument])
+  return (
+    <>
+      {config.recurring !== true && (
+        // {config.recurring !== false && (
```
eoaksnes, line 168: should this be removed? // {config.recurring !== false && (

eoaksnes (approved):
👍

PR #1129, opened 2024-01-12
fix: logs viewer scrolling

pbullhove:

1. Removed horisontal scrolling in logs, made it wrap instead. 
2. Removed "double" vertical scroll bars. So only has for the logs now, not the entire dialog as well.

How it was before: 
https://jam.dev/c/70eddea3-ded6-4e27-b95a-7911824df8b6

How it is now: 

https://jam.dev/c/a6a9fe01-4b84-446b-a559-039252ce1340


soofstad (approved):
tipp topp tommel opp

Issue #998, opened 2023-12-18
feat: improvements to datagrid component


awesthouse:
Functionality/Additional config
- [ ] Config to make the the entire grid un-editable
- [ ] Add a delete button to the band and keep all buttons visible (no change of size)
- [ ] Config to show (or not) add/delete, up/down, and pagination.
- [ ] Config to show a label above the table i.e. name of the variable for example.
- [ ] Config for showing a tooltip on the table. ❓ 
- [ ] Possibility to select a list of Rank-1 arrays attributes to be aligned as columns.
- [ ] Possibility to select a list of single primitive attributes to be aligned as single row table. ❓ 
- [ ] Possibility to copy the table to clipboard, or import data from clipboard, when clicking on the top left corner (the # sign)

Styling
- [ ] Work on sizing/resizing, test with more data
- [ ] Test out possibility of resizable columns
- [ ] Consider putting pagination controls on a separate bank below the control buttons.
- [ ] Try alignment to left, just like other items. ❓ 

Columns
- [ ] Replace headers with custom/defined column names
- [ ] Config to use letter or numbers for automatic naming of columns and rows.
- [ ] Config hide/show column with numbers (indexes)
- [ ] Config to show/hide headers for each column. ❓

eoaksnes:
Default behavior for showing add, delete etc should be by checking the dimensions. But config should be able to override this.

PR #1138, opened 2024-01-16
feat: export data as csv or copy

awesthouse:
Allows exporting data from datagrid.
 - Copy data to clipboard
 - Export as csv file

Want to export data from datagrid to use in programs like excel.

#998


soofstad (approved):
Looks good! :+1:

PR #1140, opened 2024-01-16
Feat/config tooltip

pbullhove:

image
image

### DIFF
--- a/packages/dm-core-plugins/src/form/templates/ObjectModelContainedTemplate.tsx
+++ b/packages/dm-core-plugins/src/form/templates/ObjectModelContainedTemplate.tsx
@@ -33,6 +33,7 @@ export const ObjectModelContainedTemplate = (
     uiAttribute,
     onOpen
   )
+  console.log(uiAttribute)
   return (
     <FormTemplate>
       <FormTemplate.Header>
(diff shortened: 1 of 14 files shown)


packages/dm-core-plugins/src/form/templates/ObjectModelContainedTemplate.tsx
```diff
@@ -33,6 +33,7 @@ export const ObjectModelContainedTemplate = (
     uiAttribute,
     onOpen
   )
+  console.log(uiAttribute)
```
eoaksnes, line 36: Please remove console.log :)

eoaksnes (approved):
👍

packages/dm-core-plugins/src/form/widgets/CheckboxWidget.tsx
```diff
@@ -1,18 +1,22 @@
...
+  const { value, readOnly, tooltip } = props
   return (
-    <Checkbox
-      {...props}
-      disabled={readOnly}
-      checked={value !== undefined ? value : false}
-      type='checkbox'
-      data-testid='form-checkbox'
-    />
+    <Tooltip title={tooltip ?? ''}>
```
soofstad, line 9: Are we showing an empty tooltip here?
If so, probably shouldn't

packages/dm-core-plugins/src/form/templates/ObjectModelContainedTemplate.tsx
pbullhove: thanks!

PR #1141, opened 2024-01-16
refactor: actually reverse data instead of using css

awesthouse:
- Actually reverse data in plugin before passing it to DataGrid component, reverse back before saving
- columnLabels will always be columnLabels, rowLabels will always be rowLabels
- No need to reverse data before exporting data
- Added two more examples for testing printDirection in single-primitive datagrids
- Simplifies a lot of the functionality based on printDirection

- Used to flip table only using css, meaning that HTML-markup and what was viewed was different - not great for accessibility (reading direction, tabbing happened top->down, not left->right as it should)
- columnLabels and rowLabels switching places and meaning based on printDirection was confusing to user - now it's constant. We only flip the data, not the table itself. 

#998


soofstad (approved):
Looks good! :+1:

PR #1143, opened 2024-01-17
Fix/remove log

pbullhove:

remove console.log


awesthouse (approved):
Now duplicate? This is also in the css to hide fix PR

PR #1145, opened 2024-01-17
Fix/css to hide

pbullhove:

Seems that  this solves the constant refresh when collapsing and opening of complex types in form problem.

### DIFF
--- a/packages/dm-core-plugins/src/form/templates/ArrayComplexTemplate.tsx
+++ b/packages/dm-core-plugins/src/form/templates/ArrayComplexTemplate.tsx
@@ -80,15 +80,16 @@ export const ArrayComplexTemplate = (props: TArrayTemplate) => {
             ))}
         </FormTemplate.Header.Actions>
       </FormTemplate.Header>
-      <div className={`${canExpand && isExpanded ? '' : 'hidden'}`}>
-        <FormTemplate.Content padding='px-2 pt-2'>
-          <ViewCreator
-            idReference={`${idReference}.${namePath}`}
-            onOpen={onOpen}
-            viewConfig={getExpandViewConfig(uiAttribute)}
-          />
-        </FormTemplate.Content>
-      </div>
+      <FormTemplate.Content
+        expanded={!!(canExpand && isExpanded)}
+        padding='px-2 pt-2'
+      >
+        <ViewCreator
+          idReference={`${idReference}.${namePath}`}
+          onOpen={onOpen}
+          viewConfig={getExpandViewConfig(uiAttribute)}
+        />
+      </FormTemplate.Content>
     </FormTemplate>
   )
 }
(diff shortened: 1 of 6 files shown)


packages/dm-core-plugins/src/form/templates/ArrayComplexTemplate.tsx
```diff
@@ -80,15 +80,15 @@ export const ArrayComplexTemplate = (props: TArrayTemplate) => {
             ))}
         </FormTemplate.Header.Actions>
       </FormTemplate.Header>
-      {canExpand && isExpanded && (
+      <div className={`${canExpand && isExpanded ? '' : 'hidden'}`}>
```
awesthouse, line 83: Maybe FormTemplate.Content could take in a className or hidden field so that we add the class to the wrapper that's already there? So that we don't wrap it in a extra div when it's not needed 🙂

awesthouse (changes requested):
Seems to work for me! 🎉 Just a comment on the implementation

packages/dm-core-plugins/src/form/templates/ArrayComplexTemplate.tsx
pbullhove: True, good idea!

pbullhove: will have a prop hidden: boolean

pbullhove: or expanded prop

awesthouse: Yeah, I guess default hidden and then having a visible prop also makes sense - more reflects our checks and default functionality. Just do what looks good to you 👍🏻

pbullhove: also need to make sure it does not load things that cannot be expanded .. hm

pbullhove: perhaps we should differentiate between canExpand, and isExpanded.

awesthouse: Ah, yes! Maybe we can still wrap the ViewCreator with canExpand? 'cause that doesn't change on expand/hide. We don't want to render the ViewCreator in that instance anyways

awesthouse: So pretty much what you just implemented 😆 👍🏻

pbullhove: yes!

awesthouse (approved):
🥳

PR #1146, opened 2024-01-17
fix: ui that tooltip was placed wrong

pbullhove:
The tooltip i was placed below the header row

Now it looks like this. 
image


awesthouse (approved):
LGTM!

PR #1150, opened 2024-01-17
fix: reload when expand problem for list and table

pbullhove:

Now it uses css to hide instead of conditionally rendering.


awesthouse:
Won't this load/create all views on mount and just hide with css? Say you you're showing 100 items in a list chances are hight you would only want to expand and look at 1 or a few more items. no? Won't this be bad for performance?

pbullhove:
> Won't this load/create all views on mount and just hide with css? Say you you're showing 100 items in a list, chances are high you would only want to expand and look at 1 or a few more items. Won't this be bad for performance?

Yes, you are right. I'll update with a fix for this.

packages/dm-core-plugins/src/list/ListPlugin.tsx
```diff
@@ -107,6 +107,9 @@ export const ListPlugin = (props: IUIPlugin & { config?: TListConfig }) => {
 
   const [showModal, setShowModal] = useState<boolean>(false)
   const [expanded, setExpanded] = useState<Record<string, boolean>>({})
+  const [hasBeenExpanded, setHasBeenExpanded] = useState<
+    Record<string, boolean>
+  >({})
```
soofstad, line 112: Think you can accomplish all this "hasBeenExpanded" stuff mush easier with [lazy](https://react.dev/reference/react/lazy)

soofstad (changes requested):

packages/dm-core-plugins/src/list/ListPlugin.tsx
pbullhove: Hmm nay, it seems lazy is just for whole modules, not individual items? 
I might be mistaken, but I'm not sure this will help us here.

soofstad (approved):

PR #1151, opened 2024-01-17
fix: minor ui changes to icons in table

pbullhove:

1. Jumps to last page when adding a new element. 
2. Ui Changes: 
Open icon slightly smaller. 
Chevron to the left side is now primary color (for consistency). 
Cell with chevron is not so big anymore. 

How it is now: 
image

How is was before: 
image


awesthouse (approved):
I love that our number 1 philosophy on this project is things should always up less space 😆

PR #1153, opened 2024-01-18
fix: adding tooltip

pbullhove:

Adding tooltip to "Field" so that one can have tooltips on normal input fields as well. 

image
image


soofstad (approved):
Should make a PR to EDS with this :stuck_out_tongue:

Issue #1147, opened 2024-01-17
fix: datagrid minor bugs


awesthouse:
- [x] Optional DataGrid disappears after removing the last value in form
- [x] No support for creating arrays based on dimensions if attribute was optional and undefined in form
- [x] Table by default uses too much width
- [x] Formatting "-" as a number when trying to write negative numbers creates a NaN error
- [x] Add Undo/redo button
- [x] Pagination ignores printDirection (should maybe work now with updated way of using printDirection(?))

l3abak:
The width problem is solved in the current release. No need to do more there. 
![image](https://github.com/equinor/dm-core-packages/assets/42837660/f3704393-1b81-4c28-bffa-b6ff5a985c89)

PR #1155, opened 2024-01-18
feat: support create primitive arrays based on dimensions in form

awesthouse:
- Support creating empty primitive arrays based on dimensions in PrimitiveArray field.
- Added new form to test all use-cases for dimensions and optional/required of primitive arrays in form

Haven't added support yet for creating arrays and 2D arrays in PrimitiveArrays and empty arrays with the correct dimensions and length are needed for the DataGrid component so that user can just input values. We also restrict add and delete functionality of rows and columns based on dimensions so the dimensional arrays have to be correct on creation.

#1147


soofstad (approved):
Good stuff!

PR #1157, opened 2024-01-19
Feat/datagridplugin undo

awesthouse:
- Add "undo changes" button to DataGridPlugin
- Allow writing of negative numbers in number fields

Want to be able to undo changes to data in DataGrid without refreshing.
Should be able to write negative numbers in input field without it trying to parse "-" as a number

#1147


soofstad (approved):
Looks good!

PR #1163, opened 2024-01-22
fix: explorer overflow

collinlokken:
remove fixed width on:
- explorer plugin

image
image
(scrollable horizontally and vertically)

Horizontal scroll was not possible with fixed width

### DIFF
--- a/packages/dm-core-plugins/src/data-grid/DataGridPlugin.tsx
+++ b/packages/dm-core-plugins/src/data-grid/DataGridPlugin.tsx
@@ -122,7 +122,7 @@ export function DataGridPlugin(props: IUIPlugin) {
     <Stack
       alignItems='flex-start'
       spacing={1}
-      style={{ width: '100%', height: '100%', overflow: 'auto' }}
+      style={{ width: '100%', height: '100%' }}
     >
       <DataGrid
         attributeType={attribute?.attributeType || 'string'}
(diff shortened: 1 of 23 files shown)


packages/dm-core-plugins/src/data-grid/DataGridPlugin.tsx
```diff
@@ -142,11 +142,16 @@ export function DataGridPlugin(props: IUIPlugin) {
               onClick={revertChanges}
               disabled={!isDirty}
               variant='outlined'
+              className='w-max h-max overflow-hidden'
```
soofstad, line 145: Does these buttons really need max height and width? And overflow-hidden? Seems strange...

packages/dm-core-plugins/src/grid/GridPlugin.tsx
```diff
@@ -6,7 +6,7 @@ import { TGridPluginConfig, TGridSize } from './types'
 
 const Grid = styled.div<TGridSize>`
   display: grid;
-  overflow: auto;
+  //overflow: auto;
```
soofstad, line 9:
```suggestion
```

soofstad (approved):
A lot of changes here.
We should be very careful about explicitly setting the sizes of components, will lead to less adaptable webpage.

But if this has been tested thoroughly in many different use cases (SIMPOS, screen sizes, content size, etc.) good stuff

packages/dm-core-plugins/src/data-grid/DataGridPlugin.tsx
collinlokken: The buttons have an ::after effect that overflows the button container making a scrollbar appear that should not be there. I remember we looked at that together 😃

collinlokken: w-max jsut means width: max-content, so the container gives as much space that the content needs to avoid wrapping of text etc.

PR #1164, opened 2024-01-22
fix: new list append button UI

pbullhove:

New better looking list button. 
Perhaps reuse this for table as well to make the looks more similar. 
image


awesthouse (approved):
Looks good 🙌🏻 🙌🏻🙌🏻

But, I feel like putting small UI components like this in dm-core before we are reusing it in multiple places is maybe a bit overkill. I feel like we should wait until we need it for multiple use-cases and then extract - otherwise we might be putting a lot of things in dm-core that we just forget about and don't upkeep.

Also, the Table and DataGrid "add" buttons went through multiple design iterations, so please don't change those on a whim 😅

pbullhove:
> Looks good 🙌🏻 🙌🏻🙌🏻
> 
> But, I feel like putting small UI components like this in dm-core before we are reusing it in multiple places is maybe a bit overkill. I feel like we should wait until we need it for multiple use-cases and then extract - otherwise we might be putting a lot of things in dm-core that we just forget about and don't upkeep.
> 
> Also, the Table and DataGrid "add" buttons went through multiple design iterations, so please don't change those on a whim 😅

Fair point!

PR #1169, opened 2024-01-23
fix: bug with negative page

pbullhove:

Negative page bug where having 0 items and adding one made the first page be negative. 
Fixed now for table and list.


awesthouse (approved):
LGTM 🎉

Issue #1173, opened 2024-01-23
Make performance good enough


eoaksnes:
- [x] Profile DMSS using cProfile to find bottlenecks @eoaksnes 
- [x] Cache of blueprints in frontend @soofstad 
- [x] Not invalidate cache in backend (might need to move the permission check to use case level)
- [x] Make node tree simpler? (don't fetch blueprints and meta information in tree if not needed) @eoaksnes 

- [ ] Pass down entity when available in frontend to plugins @awesthouse 
- [ ] Cache of attributes in frontend
- [ ] Blueprint lookup table, so not separate calls for invididual Blueprints?
- [ ] Investigate if can avoid some calls to DMSS API in frontend @awesthouse 

Might also check out:
- https://tanstack.com/query/latest/docs/react/overview

PR #1183, opened 2024-01-24
perf: cache getBlueprint calls

soofstad:
- Cache getBlueprint in frontend

part of #1173


eoaksnes (approved):
👍 nice

PR #1203, opened 2024-01-25
Feat/calendar bold with data

pbullhove:

Adding option in calendar to highlight dates that have data in them.
Bold and underline,. 

This is ready to be used in SIMPOS but must be released here first. 

Looks like this 
image

### DIFF
--- a/packages/dm-core/src/components/Pickers/Datepicker/Calendar.tsx
+++ b/packages/dm-core/src/components/Pickers/Datepicker/Calendar.tsx
@@ -148,15 +148,21 @@ export const Calendar = (props: CalendarProps): ReactElement => {
         </>
       ) : (
         <div className='grid grid-cols-7 gap-1'>
-          {cal.map((date, index) => (
-            <button
-              type='button'
-              onClick={() => handleDateSelection(date)}
-              aria-label={`${date.day}. ${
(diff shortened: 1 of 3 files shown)


packages/dm-core/src/components/Pickers/Datepicker/Calendar.tsx
```diff
@@ -146,15 +148,21 @@ export const Calendar = (props: CalendarProps): ReactElement => {
...
-            <button
-              type='button'
-              onClick={() => handleDateSelection(date)}
-              aria-label={`${date.day}. ${
-                Object.keys(CALENDAR_MONTHS)[date.month - 1]
-              }`}
-              className={`
-                p-1.5 w-9 rounded-full hover:bg-equinor-lightgreen hover:text-equinor-green ${
+          {cal.map((date, index) => {
+            const hasData = isDateInDatelist(
```
eoaksnes, line 152: rename to something like isHighlighted

eoaksnes (approved):
👍 smooth

packages/dm-core/src/components/Pickers/Datepicker/Calendar.tsx
pbullhove: thanks! ⭐️ done

PR #1212, opened 2024-01-26
feat: pass up selected month

pbullhove:

Welcome to my pr. 

This changes 2️⃣ things: 

1. Slight different view for the highlighted dates that have data. 

image

3. Pass up which year/month that is currently in the view. 

Earlier the only thing the Calendar/Datepicker told the parent about was which date is currently selected. 
But we need also to know what month is currently in the View. 

This is used in SIMPOS to fetch the dates for this month that have any data. 
We use this to highlight the dates with data, without having to fetch all dates for the whole platform that might have data. So we scope only down to the current view for performance reasons.

### DIFF
--- a/packages/dm-core/src/components/Pickers/Datepicker/Calendar.tsx
+++ b/packages/dm-core/src/components/Pickers/Datepicker/Calendar.tsx
@@ -170,7 +170,7 @@ export const Calendar = (props: CalendarProps): ReactElement => {
                   Object.keys(CALENDAR_MONTHS)[date.month - 1]
                 }`}
                 className={`
-                p-1.5 w-9 rounded-full justify-center hover:bg-equinor-lightgreen hover:text-equinor-green ${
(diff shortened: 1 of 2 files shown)


packages/dm-core/src/components/Pickers/Datepicker/Calendar.tsx
```diff
@@ -178,8 +183,18 @@ export const Calendar = (props: CalendarProps): ReactElement => {
                       ? ''
                       : 'text-slate-400'
                 }
-                ${isHighlighted ? 'font-bold underline' : ''}
               `}
+                style={
```
collinlokken, line 187: maybe use tailwind instead?

collinlokken (approved):
Nice! Added a minor comment

packages/dm-core/src/components/Pickers/Datepicker/Calendar.tsx
pbullhove: doesn't work, need to override button classes, so thus we need style direct

PR #1215, opened 2024-01-29
fix(job): several minor bug fixes

soofstad:
- Default endDate is 1 year ahead
- The "toggle recurring" is hidden for already created jobs
- Disable job config while a job is running
- When using a runner template with recurring jobs, make sure to updated the nested "applicationInput" field
- Don't crash on partial provided default schedule
- Throw error if plugin receives anything that is not a Job entity
- Set "triggeredBy" info in job when starting

bugs

none

### DIFF
--- a/packages/dm-core-plugins/src/job/JobControl.tsx
+++ b/packages/dm-core-plugins/src/job/JobControl.tsx
@@ -1,4 +1,5 @@
 import {
+  AuthContext,
   EBlueprint,
   ErrorResponse,
   IUIPlugin,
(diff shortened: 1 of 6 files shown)


packages/dm-core-plugins/src/job/JobControl.tsx
```diff
@@ -88,68 +117,85 @@ export const JobControl = (props: IUIPlugin) => {
...
+                >
+                  <Icon data={gear} size={24} />
+                </Button>
+              </Tooltip>
+              <TemplateMenu
+                templates={internalConfig.runnerTemplates || []}
+                onSelect={(template: TTemplate) => {
+                  dmssAPI
+                    .documentGet({ address: template.path })
+                    .then((response) => {
```
pbullhove, line 162: extract this into a hook?

pbullhove (approved):

PR #1216, opened 2024-01-29
fix: minor changes to some e2e tests and config

roar-larsen:
- Default_recipe: Added a few extra asserts to make the test more robust. Removed test.step as it was unnecessary here.
- Plugin-explorer: Added testing of deleting a entity. This also makes it possible to rerun this test locally without needing to reset app.
- Model-uncontained: The last assert was previously removed because it often failed, but it then also removed checking that we actually could update the model. The assert should be there. Added waiting for api request before proceeding which hopefully solves the previous flakiness.
- Config: Changed tracing to retain on failure as we are not doing any retries in the pipeline. Hopefully this will make it easier to debug as the video often ends to early.


soofstad (approved):
Looks good!

PR #1224, opened 2024-01-30
fix: loading ui improvements

pbullhove:

So the loading spinner was not so good. 
1. Loaded very 'quickly', making loading seem messy. 
2. Was centered, and made things seem to jump around when loading multiple plugins on the screen. 
3. Was very big, and caused a lot of attention. 

So I made the following changes. 
1. It does not appear immediately, only after loading for 300ms. This is so that components that do load very quickly will not jump around so much. 
2. Made it smaller. 
3. Made it align to top left, so less "jumping around".


soofstad (approved):
Tested! That is much better

eoaksnes (approved):

PR #1228, opened 2024-01-31
feat: add responsive grid plugin

eoaksnes:

* Add a responsive layout grid that adapts to screen size ensuring consistency across layouts
* Added logic that minimizes the sidebar on smaller screens.

https://github.com/equinor/dm-core-packages/assets/1190419/096294f4-268f-4fa9-8f75-2c07c765eda4

https://github.com/equinor/dm-core-packages/assets/1190419/0056b9c6-9b95-47fc-b94e-3a209c4e41cf

An example configuration can look like this:

```json
"initialUiRecipe": {
    "name": "Dashboard",      
    "type": "CORE:UiRecipe",
    "plugin": "@development-framework/dm-core-plugins/responsive_grid" 
    "config": {
      "type": "PLUGINS:dm-core-plugins/responsive_grid/ResponsiveGridPluginConfig",
      "rows": [
        {
          "type": "PLUGINS:dm-core-plugins/responsive_grid/RowItem",
          "columns": [
            {
              "type": "PLUGINS:dm-core-plugins/responsive_grid/ColumnItem",
              "viewConfig": {
                "type": "CORE:ReferenceViewConfig",
                "recipe": "aRecipeName",
                "scope": "...."
              },
              "size": {
                "type": "PLUGINS:dm-core-plugins/responsive_grid/ColumnSize",
                "sm": 12,
                "md": 4
              }
            },
            {
              "type": "PLUGINS:dm-core-plugins/responsive_grid/ColumnItem",
              "viewConfig": {
                "type": "CORE:ReferenceViewConfig",
                "recipe": "aRecipeName",
                "scope": "...."
              },
              "size": {
                "type": "PLUGINS:dm-core-plugins/responsive_grid/ColumnSize",
                "sm": 12,
                "md": 4
              }
            }
          ]
        },
        {
          "type": "PLUGINS:dm-core-plugins/responsive_grid/RowItem",
          "columns": [
            {
              "type": "PLUGINS:dm-core-plugins/responsive_grid/ColumnItem",
              "viewConfig": {
                "type": "CORE:ReferenceViewConfig",
                "recipe": "aRecipeName",
                "scope": "...."
              },
              "size": {
                "type": "PLUGINS:dm-core-plugins/responsive_grid/ColumnSize",
                "sm": 12,
                "md": 4
              }
            }
          ]
        }
      ]
    }
  },
```

The recipe above defines two rows, where the first row contains two columns, and the second row contains a single column.  The columns will expand on "sm" screens the whole grid (12), but on "md" screens the columns will only take up the space defined in the recipe above (4).  

Some properties of the grid.
* There are default 12 available columns to use, but we add an option to change that number of available columns through the ui recipe config if needed later on. 
* The default breakpoints for the grid are 
  * sm: 576px
  * md: 768px
  * lg:  992px
  * xl: 1200px
  * xxl: 1600px
  * xxxl: 1920px

To understand better the concept implemented take a look at these similar implementations:
- https://getbootstrap.com/docs/5.3/layout/grid/
- https://mui.com/material-ui/react-grid2/

> **NOTE**: This plugin uses https://www.npmjs.com/package/react-grid-system, but we replace it by some code that @awesthouse has later on.

### DIFF
--- a/packages/dm-core-plugins/blueprints/responsive_grid/ColumnSize.json
+++ b/packages/dm-core-plugins/blueprints/responsive_grid/ColumnSize.json
@@ -1,0 +1,47 @@
+{
+  "name": "ColumnSize",
+  "type": "dmss://system/SIMOS/Blueprint",
+  "attributes": [
+    {
+      "name": "type",
+      "type": "dmss://system/SIMOS/BlueprintAttribute",
+      "attributeType": "string"
+    },
+    {
+      "name": "xs",
+      "type": "CORE:BlueprintAttribute",
+      "attributeType": "number",
+      "optional": true
+    },
+    {
+      "name": "sm",
+      "type": "CORE:BlueprintAttribute",
+      "attributeType": "number",
+      "optional": true
+    },
+    {
+      "name": "md",
+      "type": "CORE:BlueprintAttribute",
+      "attributeType": "number",
+      "optional": true
+    },
+    {
+      "name": "lg",
+      "type": "CORE:BlueprintAttribute",
+      "attributeType": "number",
+      "optional": true
+    },
+    {
+      "name": "xl",
+      "type": "CORE:BlueprintAttribute",
+      "attributeType": "number",
+      "optional": true
+    },
+    {
+      "name": "xxl",
+      "type": "CORE:BlueprintAttribute",
+      "attributeType": "number",
+      "optional": true
+    }
+  ]
+}
--- a/packages/dm-core-plugins/src/responsive_grid/ResponsiveGridPlugin.tsx
+++ b/packages/dm-core-plugins/src/responsive_grid/ResponsiveGridPlugin.tsx
@@ -1,0 +1,54 @@
+import { IUIPlugin, Stack, ViewCreator } from '@development-framework/dm-core'
+import { Typography } from '@equinor/eds-core-react'
+import React from 'react'
+import { Col, Container, Row } from 'react-grid-system'
+import { TCol, TGridPluginConfig, TRow } from './types'
+
+const defaultConfig: TGridPluginConfig = {
+  rows: [],
+}
+
+export const ResponsiveGridPlugin = (
+  props: IUIPlugin & { config: TGridPluginConfig }
+): React.ReactElement => {
+  const { config, idReference, type, onSubmit, onChange } = props
+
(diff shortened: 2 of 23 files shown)


packages/dm-core-plugins/src/responsive_grid/ResponsiveGridPlugin.tsx
```diff
@@ -0,0 +1,54 @@
...
+        display: 'flex',
+        maxWidth: 'max-content',
+      }}
+    >
+      <Container>
+        {internalConfig.rows.map((row: TRow) => {
+          const columns = row.columns.map((col: TCol) => {
+            return (
+              <Col {...col.size}>
+                <div style={{ marginBottom: '40px' }}>
```
awesthouse, line 35: We should be using grid gap and I feel like default should be much smaller, maybe 1rem/16px? And unless <Col /> can only have one child or it's required by react-grid-system I don't think there's a need for a div inside

packages/dm-core-plugins/src/responsive_grid/ResponsiveGridPlugin.tsx
```diff
@@ -0,0 +1,54 @@
...
+                    idReference={idReference}
+                    viewConfig={col.viewConfig}
+                    onSubmit={onSubmit}
+                    onChange={onChange}
+                  />
+                </div>
+              </Col>
+            )
+          })
+          return <Row gutterWidth={row.gutterWidth || 60}>{columns}</Row>
```
awesthouse, line 49: Also think default should be smaller here. I assume gutterWidth is the same as gap.
 I think rowGap and columnGap makes sense to use and to define it as such in the plugin blueprint. Maybe there are use-cases where columnGap is different per row, but I don't think it's very common. 

Or keep gutterWidth and set default to 0, but allow user to set gutterWidth if they want different values per row + add use of column-gap and row-gap which is the default

awesthouse (approved):

packages/dm-core-plugins/blueprints/responsive_grid/ColumnSize.json
soofstad, line 1: Perhaps link to the documentation in a description, or use the description field for each attribute how this should be used

soofstad (approved):

PR #1249, opened 2024-02-01
fix: make refresh button great again (it wasn't working)

pbullhove:

So during Memoization of the UIplugin, lots of the refresh button code was deleted for some reason. 
Therefore we now make refresh button great again 🧢

1. Made it appear as it should
2. Fixed a bit on the UI: 

- Is not transparent all the time, more visible, since it is used only a few places in the application. 
- Makes refresh content when hover refresh button to highlight what is refreshed by the button. 
- A bit better border around refresh content

refreshbuttonback


soofstad (approved):
Gif in issues! :100: :1st_place_medal: 

Looks good!

PR #1266, opened 2024-02-02
fix: popovers always on top

### LINKED ISSUE #1263: fix: popover should be on top
Image

soofstad:
- Uses `react portal` along with EDS's Popover to make sure popovers are always on top and behave nicely with scroll and parent components

closes #1263


pbullhove (approved):
Great!

Issue #1268, opened 2024-02-05
fix(table): so when delete is disabled, then the (...) header should not show


pbullhove:
![Image](https://github.com/equinor/dm-core-packages/assets/37016135/f3d7580c-9fe8-435f-a959-10ae71f381ad)

PR #1269, opened 2024-02-05
fix: correct cells based on enabled functionality + pagination styling

awesthouse:
- Show correct amount of cells for functionality based on config passed so that cell amount in header matches amount in rows
- Hide additional cells if there is no use (only 1 variant, delete is also disabled)
- Fixes pagination styling bug in pic below
Skjermbilde 2024-02-05 kl  14 49 02

There are discrepancies in amount of header-cells and row-cells because we add based on functionality config. They should always be the same amount, and the content should vary based on config.

#1268

### DIFF
--- a/packages/dm-core/src/components/Table/Table.tsx
+++ b/packages/dm-core/src/components/Table/Table.tsx
@@ -67,6 +67,9 @@ export function Table(props: TableProps) {
           (variant: TTableVariant) => variant.name === tableVariant
         ).functionality
 
+  const showAdditionalCell =
+    config.variant?.length === 2 || functionalityConfig.delete
+
   /**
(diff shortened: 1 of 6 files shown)


packages/dm-core/src/components/Table/Table.tsx
```diff
@@ -118,6 +122,7 @@ export function Table(props: TableProps) {
             <TableHead
               config={config}
               tableVariant={tableVariant}
+              showAdditionalCell={showAdditionalCell}
```
pbullhove, line 125: perhaps a more descriptive name? 
what kind of cell is this? Functionality cell?

pbullhove (approved):
Looks good. Just want a different name for a variable, else great!

packages/dm-core/src/components/Table/Table.tsx
awesthouse: I agree 🎉

PR #1270, opened 2024-02-05
feat: Spoooky PR! ☠️☠️ Adding skeletons to table ☠️☠️

pbullhove:

Because of long response times we need to notify user that something is happening. 

1. Added loading skeletons on table load ( 5 rows)
2. Added loading skeleton on add row ( 1 row at the end) 
3. Added loading skeleton on delete row (1 row where the deleting row was located)
4. Disabled add / delete buttons when adding or deleting, as table does not support multiple adds/deletes at the same time. 

Looks like this: 
tableSkeletons


awesthouse (approved):
👏🏻 👏🏻 👏🏻

PR #1273, opened 2024-02-06
fix: list scroll when it should not

pbullhove:

List had an issue where it scrolled horisontally it should not. 

Fixed in this PR. 
image


awesthouse (approved):
💯

PR #1274, opened 2024-02-06
fix: table expand bug

pbullhove:
image

Was like this: 
image


awesthouse (approved):
Seems to do the trick! Weird of eds to set height in px on datacells anyways

PR #1287, opened 2024-02-09
fix: no scroll bar on pagination in list

pbullhove:

List plugin has scroll bar on pagination. 
This PR removes this.


awesthouse (approved):
💯

Issue #1264, opened 2024-02-02
fix: Not auto-load binary in form when loading the form


pbullhove:
For example: 

![Image](https://github.com/equinor/dm-core-packages/assets/37016135/7b58fae6-ccd2-45d1-9495-c694e955ea79)

PR #1291, opened 2024-02-12
Fix/dont autofetch binary

awesthouse:
- Don't automatically download files we don't have a preview for (including binary files)
- Some refactoring of MediaViewer componens
- Helper function to create synthetic download events, extracted from DataGrid

Unnecessary downloading of potentially big files that can't even be previewed. 

#1264


soofstad (approved):
Looks good!

Issue #1223, opened 2024-01-30
fix(mediaviewer): meta defaults


eoaksnes:
- [x] don't show the "Description" label in the meta box
- [x] configure show "info" (i) button
- [x] don't show title in meta should be "figure" not "Meta"

PR #1296, opened 2024-02-13
fix: hide meta when disabled

awesthouse:
- New config: showDescription
- Show info-popover when showMeta OR showDescription is enabled
- Use filename as title if label is not provided 
- Hide description label if description is not provided

Want to actually hide meta

#1223


pbullhove (approved):
Ser bra ut dette!

PR #1298, opened 2024-02-13
add meta and publish plugins

### LINKED ISSUE #1148: feat: publish plugin
Tasks:
- [x] add copy/move/link functionality
- [ ] add metadata
- [ ] confirm dialog where can set metadata

We need a UIPlugin that could be configured to copy/move/link one entity from one place to another place. This is often used, for example, to publish the result of an analysis to an official repository. Or to commit changes to a master model.

## Add functions inside core to handle copy or link 

Preliminary arguments:
- source: with scope supporting relative path
- destination: 
  - link mode: where to push the link for the copied entity, if action is copy and link
  - copy mode: where to copy the entity from source, with scope supporting relative path
- mode: copy or link

There could also be two functions instead of the mode. 

## Add publish plugin

Create a new plugin that wraps the entity that it wants to publish. The new publish plugin should have a header and a button to publish. 

The configuration should be:

```json
{ 
  "destination": "/some/path",
  "links": ["$1234.some-list"],
  "archiveWrapper": true, 
  "viewConfig": { }
}
```

The archiveWrapper is an option to keep information about who published and when. If true, can put the `created_by`, `modified_by`, `created_date`, and `modified_date` ... that are published inside the [meta](https://github.com/equinor/data-modelling-storage-service/blob/master/src/SIMOS/Meta.json). 

The destination is optional, and if the destination is not given in the UI recipe, can use a picker to select a destination. If given should not be possible/visible to pick it in the web interface.

soofstad:
Adds plugins for "Publish" and "Meta".
They are both wrappers, that just show a "publish button", and "meta info" before the main entity's plugin.
Screenshot_20240214_124130
Screenshot_20240214_133534

Closes #1148

### DIFF
--- a/packages/dm-core-plugins/src/meta/MetaPlugin.tsx
+++ b/packages/dm-core-plugins/src/meta/MetaPlugin.tsx
@@ -1,0 +1,89 @@
+import {
+  EntityView,
+  IUIPlugin,
+  Loading,
+  TGenericObject,
+  TMeta,
+  useDocument,
+} from '@development-framework/dm-core'
+import { Table } from '@equinor/eds-core-react'
+import { DateTime } from 'luxon'
+
+export const MetaPlugin = (props: IUIPlugin) => {
+  const { document, isLoading, error } = useDocument<TGenericObject>(
+    props.idReference,
+    0
+  )
+
+  if (isLoading) return <Loading />
+
+  if (error) throw new Error(JSON.stringify(error, null, 2))
+
+  if (!document) return
+
+  const meta: TMeta = document._meta_
+
+  if (document.type !== 'dmss://system/Plugins/dm-core-plugins/common/Meta')
+    throw new Error(
+      'The meta plugin only supports entities of type "Meta" at this stage'
+    )
+
+  return (
+    <>
+      <Table className={'w-full'}>
+        <Table.Head>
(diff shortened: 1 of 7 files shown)


packages/dm-core-plugins/src/meta/MetaPlugin.tsx
```diff
@@ -0,0 +1,89 @@
...
+          <Table.Row>
+            <Table.Cell>Created by</Table.Cell>
+            <Table.Cell>Created time</Table.Cell>
+            <Table.Cell>Last modified by</Table.Cell>
+            <Table.Cell>Last modified time</Table.Cell>
+          </Table.Row>
+        </Table.Head>
+        <Table.Body>
+          <Table.Row>
+            <Table.Cell>{meta?.createdBy ?? 'nil'}</Table.Cell>
```
awesthouse, line 44: In TablePlugin we use "-" to show null values, feel like it's a bit less distracting

awesthouse (approved):
Looks really good!

packages/dm-core-plugins/src/meta/MetaPlugin.tsx
soofstad: Yeah thats better. Fixed!

PR #1309, opened 2024-02-15
fix: call form onChange instead of save in DataGridPlugin if passed

awesthouse:
When DataGridPlugin is a part of form/onChange function is passed to plugin, call onChange function instead of save

Form overwrites changes in DataGradPlugin if it's nested more than one time.

#1294

### DIFF
--- a/packages/dm-core-plugins/src/data-grid/DataGridPlugin.tsx
+++ b/packages/dm-core-plugins/src/data-grid/DataGridPlugin.tsx
@@ -78,16 +78,16 @@ export function DataGridPlugin(props: IUIPlugin) {
 
   function parseDataBeforeSave() {
     let modifiedData = data
-      if (config.printDirection === 'vertical') {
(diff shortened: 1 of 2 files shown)


pbullhove:
Nice job on fixing the bug. 

I'll approve this PR, for it fixes the bug. But it seems the way it works now is that you must submit twice  (both inner submit and then outer submit) for the changes to take place. 

I really don't like that. I think that in this case we must be able to have only one submit button, as a lot of users will think they have submitted and are finished when the inner submit is pressed.

pbullhove (approved):
Nice, see comment!

pbullhove:
@awesthouse Super thinking, thanks for a great answer. 
Given what you say, I agree it's not feasible to do now with the current priorities. 
Update seems better in this case! 

Trying to think about where I have seen similar usecases. 
All of excel/sheets have autosave. 

One way to signify to the user that this is dirty-state is to use the blue color as we do in the other form cells. Would that work? I think that would solve most of the user-confusion.

Issue #1252, opened 2024-02-02
fix(job): UI bugs cron-jobs


roar-larsen:
- [x] Simple view: Changing "interval" makes time show in cron-format which is not wanted and mismatching the dropdown list. Should show in same format as dropdown list.
![Image](https://github.com/equinor/dm-core-packages/assets/38418577/864c7915-20bb-43af-ac07-6eb01b5670bf)

- [x] Simple view: You should not be able to cross values out or type values yourself, this will opens for invalid cron-syntax.

- [x] Advanced view: If you type invalid cron-syntax and run, the plugin crashes. Perhaps difficult to validate cron-syntax in frontend, but at least a more user friendly error needs to be given.

![Image](https://github.com/equinor/dm-core-packages/assets/38418577/d2584d9d-bff1-401f-aa62-e8afc6b5bb0e)

- [x] Inform that job scheduling is always in UTC-timing?

- [x] Advanced and simple button changes position when clicked

PR #1314, opened 2024-02-16
fix(job): a few bugs in scheduler ui

### LINKED ISSUE #1303: refactor(job-create): remove time from periodic jobs

soofstad:
- Can no longer set time for recurring jobs "from" "to"
- Plugin no longer crashes if trying to start with invalid cron syntax
- Added text to inform  that times are in UTC
- Styling so that "Simple" and "Advanced" button don't shift around

bugs

closes #1303 
a few from #1252

### DIFF
--- a/packages/dm-core-plugins/src/job/JobControl.tsx
+++ b/packages/dm-core-plugins/src/job/JobControl.tsx
@@ -19,7 +19,9 @@ import { Button, Chip, Icon, Tooltip } from '@equinor/eds-core-react'
 import { gear } from '@equinor/eds-icons'
 import { AxiosError } from 'axios'
 import _ from 'lodash'
+import { DateTime } from 'luxon'
 import { useContext, useEffect, useState } from 'react'
+import { IAuthContext } from 'react-oauth2-code-pkce'
 import { toast } from 'react-toastify'
 import {
   ConfigureRecurring,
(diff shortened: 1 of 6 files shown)


packages/dm-core-plugins/src/job/JobControl.tsx
```diff
@@ -75,8 +77,19 @@ export const JobControl = (props: IUIPlugin) => {
     if (!jobEntity) return
     jobEntity.triggeredBy =
       tokenData?.preferred_username || tokenData?.name || 'Anonymous'
-    // @ts-ignore
-    if (asCronJob) jobEntity.schedule = schedule
+    if (asCronJob) {
+      // @ts-ignore
```
awesthouse, line 81: I agree with linting here - why are we updating state in a non-immutable way like this? Goes for both triggeredBy and schedule

packages/dm-core-plugins/src/job/SimpleJobControlButtons.tsx
```diff
@@ -124,3 +134,17 @@ export const CompletedButton = (props: {
...
+export const ErrorButton = (props: { error: ErrorResponse }) => {
+  return (
+    <Button
+      style={{
+        backgroundColor: loadingColor,
+        cursor: 'not-allowed',
+      }}
+      variant='contained_icon'
+    >
+      <Icon data={error_outlined} style={{ color: 'white' }} />
```
awesthouse, line 147: <Icon /> also has a color prop if you want to use that 🙂 
`<Icon data={error_outlined} color="white" />`

awesthouse (approved):
Language and text seems much better and clearer 🙂 Looks good! And looks like we were already doing the state mutating before this pr, but I don't see a reason for doing it this way, would be good to fix it if there isn't a good reason

packages/dm-core-plugins/src/job/JobControl.tsx
soofstad: The linter did not complain, its typescript that is not able to see that the job is a recurring job.
However, no reason to mutate a state object neither :slightly_smiling_face:

PR #1321, opened 2024-02-20
fix(job): replace eds autocomplete with the new NativeSelect

### LINKED ISSUE #1252: fix(job): UI bugs cron-jobs

soofstad:
- Should never have been an AutoComplete, but EDS had no Select at the time. Now it does!
- Remove non-existing field from Job recipe

closes #1252

### DIFF
--- a/example/app/data/DemoDataSource/recipes/job.recipe.json
+++ b/example/app/data/DemoDataSource/recipes/job.recipe.json
@@ -43,12 +43,6 @@
                     "widget": "DateTimeWidget"
                   },
                   {
-                    "name": "ended",
-                    "type": "PLUGINS:dm-core-plugins/form/fields/StringField",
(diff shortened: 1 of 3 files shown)


awesthouse (approved):
LGTM!

PR #1332, opened 2024-02-27
feat: don't use index for address of item when 'ensureUID'

### LINKED ISSUE #1308: feat(list): test operating on a list where the elements are addressed by id (ensureUID attribute)
Addressing elements in a list by index is flakey. As the the list may change.

- [ ] Create a list of jobs, where the elements all has a _id set by "ensureUID"
- [ ] Create multiple jobs, start them, delete some, and make sure the original job is still the one operated on (saved/updated by the jobAPI).

soofstad:
- Items in lists (useList hook), now has an `idReference`. 
  - Will be set to `_id=...` if "ensureUID", else use index

depends on:
- https://github.com/equinor/data-modelling-storage-service/pull/807

closes #1308

### DIFF
--- a/example/app/data/DemoDataSource/plugins/list/task_list/blueprints/TaskList.blueprint.json
+++ b/example/app/data/DemoDataSource/plugins/list/task_list/blueprints/TaskList.blueprint.json
@@ -17,7 +17,8 @@
       "type": "CORE:BlueprintAttribute",
       "attributeType": "./Task",
       "contained": true,
-      "dimensions": "*"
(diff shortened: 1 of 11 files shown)


awesthouse (approved):
Looks really good! Also tested some locally 🙂 

I think we can deprecate and remove the index attribute in list-items - as far as I can tell it's only used in the view label for onOpen in ListPlugin and we might as well use the print-index there. It would probably be nice to do it in a separate pr tough, so that we can test both things separately, so there's no need to add it at this point.

PR #1354, opened 2024-03-04
fix(datagrid): pagination should be high enough for numbers to fit

pbullhove:
# Now looks like this: 
image

#  Before it looked like this: 
image


awesthouse (approved):
👍🏻

PR #1369, opened 2024-03-06
fix: make datagrid clear button 

pbullhove:

1. added clear button, and clearing functionality
2. fixed a datagrid bug where one could not submit empty datagrid
3. added tooltips to the small buttons. 
4. made empty datagrid be full width 
5. made action row have border on the left side 

image

image

### DIFF
--- a/packages/dm-core/src/hooks/useJob.tsx
+++ b/packages/dm-core/src/hooks/useJob.tsx
@@ -132,6 +132,7 @@ export function useJob(entityId?: string, jobId?: string): IUseJob {
       return null
     }
     setIsLoading(true)
+    console.log(' --- Peter --- ')
     return dmJobApi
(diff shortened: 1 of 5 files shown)


packages/dm-core/src/hooks/useJob.tsx
```diff
@@ -132,6 +132,7 @@ export function useJob(entityId?: string, jobId?: string): IUseJob {
       return null
     }
     setIsLoading(true)
+    console.log(' --- Peter --- ')
```
eoaksnes, line 135: image

eoaksnes (approved):
👍  Just remove the console.log :)

packages/dm-core/src/hooks/useJob.tsx
pbullhove: aw ye

PR #1371, opened 2024-03-07
refactor: merge all contexts into one

soofstad:
This is a big one. Touches nearly every single file....

Merge all our Contexts into one.


pbullhove (approved):
Sweet, seems more clean and consistent now!

PR #1378, opened 2024-03-11
Fix/sidebar navigate trigger reload

### LINKED ISSUE #1374: Sidebar select item should remount / update items
Optimistic load? / lazy so that we don't loose performance, but instead update when needed

soofstad:

- Sidebar plugin now refreshes the content when navigating

- Navigating using the "Sidebar"-plugin should refresh "sub-pages", and not show outdated content.

fixes #1374


example/app/data/DemoDataSource/plugins/view_selector/roles_school_example/school.entity.json
```diff
@@ -1,4 +1,5 @@
 {
+  "_id": "hogwars-school-meta",
```
pbullhove, line 2: typo: hogwarts (with t)

also, perhaps rename to `hogwarts-school-district`, since it's not a meta type?

pbullhove: if it's used in a test where it tests _meta_ then perhaps it's okay.

pbullhove (approved):

PR #1379, opened 2024-03-11
Refactor/form-cleanup

awesthouse:
**Form cleanup**
- A lot of components get their own folder and components only used in that component are moved from "general" folders to where they are used
- Remove some unnecessary mapping and and spread operators
- Utilities cleanup
- Split FormTemplate/ComplextAttributeTemplate so that it keeps 1 component per file

**Renaming**
- FormTemplate -> ComplexAttributeTemplate
- AttributeField -> AttributeFieldSelector

Just some cleanup 🙂


pbullhove (approved):
Looks pretty good. 
I agree on the name changes, i think they are fine.

PR #1380, opened 2024-03-11
Feat/job control styling improvements

pbullhove:

Looks better. 

image

image
image

Was ugly.

### DIFF
--- a/packages/dm-core-plugins/src/job/CronJob.tsx
+++ b/packages/dm-core-plugins/src/job/CronJob.tsx
@@ -74,123 +74,119 @@ export function ConfigureSchedule(props: {
 
   return (
     <div>
-      <div>
-        {isRegistered && (
-          <Typography color='primary'>The job is scheduled</Typography>
-        )}
-        <DateRangePicker
-          setDateRange={(dateRange) =>
-            setSchedule({
-              ...schedule,
-              startDate: dateRange.startDate,
-              endDate: dateRange.endDate,
-            })
-          }
-          value={{ startDate: schedule.startDate, endDate: schedule.endDate }}
-        />
-        <InputWrapper>
-          {showAdvanced ? (
-            <div className={'flex-col'}>
-              <TextField
-                style={{ maxWidth: '400px' }}
-                unit='cron'
-                id='advanced-schedule-syntax'
-                type='text'
-                defaultValue={schedule.cron}
-                onChange={(event: ChangeEvent<HTMLInputElement>) =>
-                  setSchedule({
-                    ...schedule,
-                    cron: event?.target.value,
-                  })
-                }
-                label='Enter explicit cron syntax'
-                helperText='minute hour day(month) month day(week)'
-              />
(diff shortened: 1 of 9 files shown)


packages/dm-core-plugins/src/job/CronJob.tsx
```diff
@@ -73,131 +73,120 @@ export function ConfigureSchedule(props: {
...
+            ...schedule,
+            startDate: dateRange.startDate,
+            endDate: dateRange.endDate,
+          })
+        }
+        value={{ startDate: schedule.startDate, endDate: schedule.endDate }}
+      />
+      <InputWrapper>
+        {showAdvanced ? (
+          <div className={'flex-col'}>
```
awesthouse, line 92: flex-col wont have any effect on a div that doesn't have flex display

packages/dm-core-plugins/src/job/CronJob.tsx
```diff
@@ -73,131 +73,120 @@ export function ConfigureSchedule(props: {
...
+              onChange={(event: ChangeEvent<HTMLInputElement>) =>
+                setSchedule({
+                  ...schedule,
+                  cron: event?.target.value,
+                })
+              }
+              label='Enter explicit cron syntax'
+              helperText='minute hour day(month) month day(week)'
+            />
+            <small style={{ color: 'red' }}>
```
awesthouse, line 108: color: red doesn't have enough contrast - should at least use tokens.colors.interactive.danger__text.hex

packages/dm-core-plugins/src/table/TablePlugin.tsx
```diff
@@ -28,21 +28,23 @@ export const TablePlugin = (props: IUIPlugin) => {
...
-      onOpen={props.onOpen}
-      reloadData={reloadData}
-      removeItem={removeItem}
-      saveTable={save}
-      setDirtyState={setDirtyState}
-      setItems={setItems}
-      updateItem={updateItem}
-      type={props.type}
-    />
+    <div className='dm-plugin-padding'>
```
awesthouse, line 31: Maybe triple check that this maybe shouldn't have w-full as well - I think it works most places because we use flex pretty much everywhere that stretches children, but I don't know how this will look in grid if say the table is narrower in width than parent 🙂

awesthouse (approved):
Really like the changes 👏🏻 👏🏻 👏🏻 

Just commented on a few things that isn't necessarily changes you made, but if you feel like fixing them it would be great 👍🏻

packages/dm-core-plugins/src/job/CronJob.tsx
pbullhove: aright, done!

packages/dm-core-plugins/src/table/TablePlugin.tsx
pbullhove: aha, I think that solves the table not being 100% on the botton I asked about earlier.

PR #1383, opened 2024-03-11
fix(data-grid): disable clear button when rows are not editable

pbullhove:

Only show clear button when rows are editable


awesthouse (approved):
LGTM!

PR #1387, opened 2024-03-12
fix: remove dependency + call function so that it doesn't create loop

awesthouse:
Fix clearTable in DataGrid


pbullhove (approved):
smart

PR #1396, opened 2024-03-14
refactor: tailwind build cleanup

awesthouse:
- Removes unused reset stylesheet (dm-core-plugins/build.css)
- Use input file for tailwind build and create 1 single build file per package + put output file directly in dist folder
- Move custom css from dm-core to dm-core packages
- Don't export css in index files for packages
- Update import in example app to use stylesheet from dm-core-plugins

- Unnecessary exports and copying of build files
- Unused stylesheets
- Styles should be in dm-core-plugins


pbullhove (approved):
Goodie!

PR #1403, opened 2024-03-18
fix: add flex-row class after custom layout class

awesthouse:
Add flex-row class after custom layout class

tailwind classes are bundled after custom css so it doesn't overwrite custom css which is flex-direction column


soofstad (approved):
:+1:

PR #1408, opened 2024-03-19
Docs/header plugin test- just a draft of the docs to see if it works. 

pbullhove:

Just adding some draft docs. 
This must be published for the tsDocs to find it and publish it in the docosaurus.


awesthouse (approved):
Nice! 👍🏻 

1. If there are cases where we want to use more than one image, maybe we should have a static or image folder to put them in - just to get some structure
2. Should use jpeg for images to reduce file size

PR #1426, opened 2024-05-15
Feat/refresh application

### LINKED ISSUE #1410: Reset UI (lookup table)
UI for resetting application (recreating recipe lookup table)

Add the admin icon to the header where you can do the reset.

Create the reset UI as plugin, so that it can be-reused later on.

soofstad:

- Adds a button to the "User" popup to reset application

closes #1410


packages/dm-core-plugins/src/header/components/UserInfoDialog.tsx
```diff
@@ -147,6 +147,25 @@ export const UserInfoDialog = (props: UserInfoDialogProps) => {
               >
                 Create API-Key
               </Button>
+              <Button
```
awesthouse, line 150: Is this supposed to be available for all user-roles? Feel like it should be somewhat restricted

awesthouse (approved):
👍🏻 Just one comment

packages/dm-core-plugins/src/header/components/UserInfoDialog.tsx
soofstad: Agreed. Updated HeaderConfig.json to now include "adminRole". If the user has that role, the "admin menu" is shown in the top bar :+1:

PR #1430, opened 2024-05-21
feat(yaml): raw json edit

### LINKED ISSUE #1411: feat(json-edit): new plugin to view and edit raw json
a UI Plugin to view, edit, and save documents in JSON format. Similar to what Mongodb UI provides. We can offer validity check of the document when the use hit save as well so it would not be possible to save an invalid document.

soofstad:
- Implements a _simple_ "raw edit json mode" to the yaml-plugin 
Screenshot_20240521_134044

 Its very basic and limited. But i think it does the job, and avoids pulling in a 100MB editor NPM dependency

closes #1411

### DIFF
--- a/packages/dm-core-plugins/src/yaml/YamlPlugin.tsx
+++ b/packages/dm-core-plugins/src/yaml/YamlPlugin.tsx
@@ -1,16 +1,34 @@
 import {
+  ErrorGroup,
   IUIPlugin,
   Loading,
   TGenericObject,
   useDocument,
 } from '@development-framework/dm-core'
-import { Button, Input, Label } from '@equinor/eds-core-react'
+import { Button, Input, Label, Switch } from '@equinor/eds-core-react'
 import DOMPurify from 'dompurify'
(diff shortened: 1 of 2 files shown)


awesthouse (approved):
Nice! This is definitely good enough to close the issue

Some suggestions though:
- I understand the purpose of the 2 save buttons, but having 2 save buttons for the same thing means that 1 of them wont really have a natural connection to the element it controls. Maybe we can look at some type of sticky bar at the bottom with a save button that is always visible?
- An editor like [@monaco-editor/react](https://www.npmjs.com/package/@monaco-editor/react) isn't that bad in size if we want to use a package
- Maybe it would be better to allow user to switch between json-view and yaml-view?
  - Edit available in both json and yaml language
  - copy buttons based on which view you've selected that are more naturally integrated and take up less space
  - Name of plugin should probably be more generic - like EditorPlugin

I'm fully back this week so I have more time to work on stuff like this if you want me to - at least maybe a design improvement would be nice 🙂 I don't think it would take too long

PR #1439, opened 2024-05-27
Feat/yaml plugin

awesthouse:

**useDocument**
- Export setError function so that error can be reset from plugin
- Add parameter `throwError` to updateDocument function that decides if error should be actually thrown in updateDocument so that it's easier to handle errors in plugins  when needed

**YamlPlugin**
- Add language config. Expects a list of languages (json, yaml) so that user can decide both order and enabled languages in one go
- Add editable config. Default to true, but should be able to disable edit functionality.

- Can now both view, edit and copy code  in yaml and json respectively
- Sticky sidebar with buttons on side, scroll in code wrapper
- Some error handling

- Want to combine YAML plugin and JSON plugin to one + expand editor functionality
- Tried to reduce the feeling of floating buttons
- View was only available in YAML, edit was only available in JSON

Skjermbilde 2024-05-27 kl  15 20 24

NOTE: In the future we have a plan to rename plugin to something like InspectPlugin

### DIFF
--- a/packages/dm-core-plugins/src/yaml/YamlPlugin.tsx
+++ b/packages/dm-core-plugins/src/yaml/YamlPlugin.tsx
@@ -8,7 +8,7 @@ import { Icon, Popover, TextField, Tooltip } from '@equinor/eds-core-react'
 import { close, copy, edit, filter_alt, save } from '@equinor/eds-icons'
 import DOMPurify from 'dompurify'
 import hljs from 'highlight.js'
-import { ChangeEvent, useEffect, useRef, useState } from 'react'
+import { ChangeEvent, useMemo, useRef, useState } from 'react'
 import { toast } from 'react-toastify'
 import YAML from 'yaml'
 import { ActionButton, ActionsWrapper, CodeContainer } from './styles'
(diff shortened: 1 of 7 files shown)


soofstad (changes requested):
Very nice! Looks great.
Two things:
 - It does not look very nice when you get a scroll bar to the right of the button. Perhaps move the button wrapper and the code into one div that gets a scroll bar instead. And use relative positioning to have them follow the scroll.
 - IMO we should not throw error inside "useDocument", even if it's optional. Why was this needed now?

packages/dm-core-plugins/src/yaml/YamlPlugin.tsx
```diff
@@ -1,173 +1,181 @@
...
-
-  & .hljs-string {
-    color: #a5ff90;
-  }
+export const YamlPlugin = (props: YamlPluginProps) => {
+  const { idReference, config: userConfig } = props
+  const config = { ...defaultConfig, ...userConfig }
+  const [depth, setDepth] = useState(0)
+  const [asYAML, setAsYAML] = useState<string>('')
+  const [asJSON, setAsJSON] = useState<string>('')
```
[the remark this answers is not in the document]
awesthouse: Yes! Nice!

packages/dm-core/src/hooks/useDocument.tsx
```diff
@@ -51,7 +52,8 @@ interface IUseDocumentReturnType<T> {
 export function useDocument<T>(
   idReference: string,
   depth?: number | undefined,
-  notify: boolean = true
+  notify: boolean = true,
+  throwError: boolean = false
```
soofstad, line 56: IMO, "throwError" should be moved into "updateDocument". As not to be confused with throwing error on getting the document.

soofstad (approved):

packages/dm-core/src/hooks/useDocument.tsx
awesthouse: Yes, agree! ✅

PR #1440, opened 2024-05-28
Docs/more plugin docs

soofstad:
- fix the blueprint hierarchy plugin
- add docs to
  - explorer
  - blueprint hierarchy
  - header
- delete deprecated pdf plugin


awesthouse (approved):
Nice! 💯  I was curious if the pdf plugin was a bit redundant 

Even tough we're removing docs from the actual rpm package, I do think for our own sake we should try and reduce file sizes as much as possible so I think it would be better to use JPEGs instead of PNGs. At least on mac OS you can configure screenshot filetype to use JPEG and windows snipping tool also lets use save to JPEG so it shouldn't be too difficult to switch

PR #1441, opened 2024-05-29
feat(job): configurable title

### LINKED ISSUE #1418: Job runner ui config for showing a header
So add to the job runner config an option to add a header/title thing.

soofstad:
Adds a optional "title" to jobPluginConfig and UI

Be able to describe what the "play button" does

closes #1418

### DIFF
--- a/example/app/data/DemoDataSource/recipes/plugins/job/jobContainer.recipe.json
+++ b/example/app/data/DemoDataSource/recipes/plugins/job/jobContainer.recipe.json
@@ -6,6 +6,7 @@
     "type": "CORE:UiRecipe",
     "config": {
       "type": "PLUGINS:dm-core-plugins/job/CreateConfig",
+      "title": "Run a reverse description job",
       "hideLogs": true,
       "jobTargetAddress": ".job",
       "jobTemplates": [
--- a/example/app/data/DemoDataSource/recipes/recurringJob.recipe.json
+++ b/example/app/data/DemoDataSource/recipes/recurringJob.recipe.json
@@ -108,6 +108,7 @@
               "type": "CORE:UiRecipe",
               "config": {
                 "type": "PLUGINS:dm-core-plugins/job/ControlConfig",
+                "title": "Schedule an example recurring job",
                 "runnerTemplates": [
                   {
                     "type": "PLUGINS:dm-core-plugins/common/Template",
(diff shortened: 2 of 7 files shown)


awesthouse (approved):
Nice! 💯 

I added a commit with some css cleanup if that's okay. dm-plugin-wrapper class gives us the opportunity to create spacing around plugins so combining it with the plugin wrapper created duplicate padding when it wasn't in Grid (like the recurring job example) so it might not have been that noticeable. I just reintroduced It and cleaned up some unnecessary css classes. Also used the EDS Typography component. For text it's fine to use tailwind/nothing, but headings are basically unstyled after tailwind reset, so we either need to use EDS typography (which is probably what we should do) or tailwind classes 🙂

soofstad:
> Nice! 💯
> 
> I added a commit with some css cleanup if that's okay. dm-plugin-wrapper class gives us the opportunity to create spacing around plugins so combining it with the plugin wrapper created duplicate padding when it wasn't in Grid (like the recurring job example) so it might not have been that noticeable. I just reintroduced It and cleaned up some unnecessary css classes. Also used the EDS Typography component. For text it's fine to use tailwind/nothing, but headings are basically unstyled after tailwind reset, so we either need to use EDS typography (which is probably what we should do) or tailwind classes 🙂

Just a few questions :slightly_smiling_face: 

If 'dm-plugin-padding' is a "common plugin padding class", why can you not combine it with other classNames?

Also, Equinor Typograpy is fine, but what is the big benefit? Would it not be better to just let the imported equinor-fonts in index.html set the fonts? Would make it easier to change later also. The result is the same anyhow

awesthouse:
> Just a few questions 🙂
> 
> If 'dm-plugin-padding' is a "common plugin padding class", why can you not combine it with other classNames?
> 
> Also, Equinor Typograpy is fine, but what is the big benefit? Would it not be better to just let the imported equinor-fonts in index.html set the fonts? Would make it easier to change later also. The result is the same anyhow

It's meant to help us create padding AROUND the plugin. Before I put it back it was on the job-wrapper that has a border and background which meant that the job-wrapper got more padding than we wanted. I can look into using margin instead, but you can still end up with the same problem of giving margin to an element you don't want to. Most plugins have a wrapper that isn't styled like this one, so it's usually not a problem 🙂 

The font will be Equinor, but it won't have any styling. Tailwind reset removes all heading sizes and margins (and for good reason 'cause they usually aren't that great) meaning that we either have to use Equinor component or tailwind typography classes. 
Skjermbilde 2024-05-30 kl  09 45 24

PR #1453, opened 2024-06-17
refactor: remove tailwind

awesthouse:

**both packages and example application**
- Switch out tw classNames with layout component and inline styles where necessary
- Remove tailwind, autoprefixer and postcss from projectsa
- Cleaned up use of titles and aria-labels for buttons and icon-buttons
- Some additional aria-attribute improvements
- Update styles after removal of tailwind reset
- Remove unnecessary div wrappers, elements and styles/classNames
- Minor general improvements of semantic HTML and roles

**dm-core**
- Stack as layout component in dm-core to avoid dependencies to dm-core-plugins
- Some style cleanup in ACL components
- Minor styled-components global styles

**dm-core-plugins**
- Expanded Stack functionality to handle more use-cases
- Use of singular button in ComplexAttributeTitle, not multiple buttons with same onClick
- Hyperlinkwidget redesign

- Testing easier ways to include css in projects for end-users/no additional import of stylesheets
- Less dependencies
- Plugins can "stand on their own" without external style dependencies and resets


awesthouse:
@soofstad Thank you reviewing! It was hard to split in to multiple PR's because it's so connected and it was hard to see if everything was correct until I actually stripped everything away. It's also impossible to read! When you edit wrappers like this it's almost impossible to see the actual changes, but I tried to write the PR description to contain everything that's been done.

I tried not using unnecessary stacks, maybe I missed a few things that can be grouped with a simple div - will definitely take a look!

Truuee, our track record isn't that great, but I have tested a lot! Will pay close attention and test more. I have a PR coming in SIMPOS and will update dm-create-app as well 🙂

PR #1467, opened 2024-12-13
Refactor/update packages

awesthouse:
- Need to update packages
- Package build broken due to linting errors
- docker-compose not working properly in github runner
- e2e tests not working properly

- chore: update packages
- fix: linting/biome errors and build errors
- fix: switch docker-compose with docker compose in integration test job
- fix: e2e test fixes

**Most of the changes are**
- added "type" before type imports
- styled-components filter props, either by filtering with config or prefixing with $
- use of Number.parseFloat and Number.parseInt instead of global parseFloat and parseInt


.github/workflows/integration-tests.yaml
```diff
@@ -56,27 +56,27 @@ jobs:
         working-directory: example/
 
       - name: Pull docker images
-        run: docker-compose pull
```
awesthouse, line 59: [https://github.com/orgs/community/discussions/116610#discussioncomment-8997411](https://github.com/orgs/community/discussions/116610#discussioncomment-8997411)

Might be a better solution for this, but this does the trick. Looks like this has been broken for a while now

PR #1472, opened 2025-01-20
Fix/datagrid bugs

### LINKED ISSUE #1463: bug: arrays of integer type are treated as strings in a default DataGrid
#### Expected Behaviour
Numbers are treated as strings by input, but should be parsed back to number before saving to db

#### Actual Behaviour
Arrays of integer type are treated as strings in a default DataGrid (no specific UI recipe). lineNumbers and serviceLife are arrays of integers as defined in Blueprint, but this is how they are stored in MongoDB after passing through the DataGrid.

### LINKED ISSUE #1464: fix: copy from excel adds additional line (cell) that fail validation if datagrid attributes are numbers

### LINKED ISSUE #1465: fix:  add rows button should disappear when "adjustableRows" is set to false
#### Expected Behaviour
Add rows button should disappear when "adjustableRows" is set to false

#### Actual Behaviour
Still showing

awesthouse:
- fix: integers are treated as strings
- fix: copy from excel adds line to data object
- fix: add rows button still enabled tough adjustableRows is set to
false
- fix: export functionality doesn't work when including row labels
- fix: missing labels  for some use-cases
- chore: add datagrid e2e tests
- chore: add test case for numeric (integer) type

- Refactored the way we create labels. 
- Removed possibility to merge custom and predefined labels as it over-complicated the functionality and the usage is pretty low/non-existent 

Datagrid seems to have quite a few bugs - need some refactoring/simplifying to reduce complexity 

Closes #1463 
Closes #1464 
Closes #1465


collinlokken (approved):
Nice!

Issue #1482, opened 2025-03-06
fix: media viewer multi bug


awesthouse:
- Previews not loading correctly
- Previews not supporting width and height config properly
- Hard to configure preview size with progressive sizing (like in use of grid and stack)

PR #1483, opened 2025-03-06
Fix/media viewer

awesthouse:
- refactor: clean up code, split components
- fix: image tag loading before image blob is fetched - meaning that depending on internet speed you see a broken image tag before image is loaded
- fix: all previews support height and width config
  - width and height now working properly on pdf preview
  - added "fill" config that allows you to fill width, height or both

- Previews not loading correctly
- Previews not supporting width and height config properly
- Hard to configure preview size with progressive sizing (like in use of grid and stack)

#1482


eoaksnes (approved):
Good code changes! Big improvement :)

